PR #18, opened 2021-09-18
Additional db seeds

obozhinov:


db/seed/beneficiary.seed.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  const coordinatorFromDb = await prisma.coordinator.findFirst()
+  console.log(coordinatorFromDb)
+
+  if (!coordinatorFromDb) {
+    throw new Error('No coordinator')
+  }
+
+  var result= new Array();
```
kachar, line 35: Usage of `const` is advised over `var` or `let` to provide immutability. There's no problem adding items to the array as the type of the variable is not being changed.

```suggestion
  const result= []
```

db/seed/beneficiary.seed.ts
```diff
@@ -0,0 +1,45 @@
+import { PrismaClient, PersonRelation } from '@prisma/client'
+const prisma = new PrismaClient()
+
+export async function beneficiarySeed() {
+  console.log('Beneficiary seed')
+
+  const persons = await prisma.person.findMany({ where: { newsletter: true } })
```
kachar, line 7: What's the correlation between people that are subscribed to newsletter and beneficiary seed?

db/seed/person.seed.ts
```diff
@@ -0,0 +1,21 @@
+import { PrismaClient } from '@prisma/client'
+const prisma = new PrismaClient()
+
+export async function personSeed() {
+  console.log('Persons seed')
+
+  const insert = await prisma.person.createMany({
+    data: [
+      { firstName: 'Sofia', lastName: 'Vicheva', email: 'sofia@abv.bg', phone: '+359888910910', company: 'ABC', address: 'Stefan Karadja 5, Sofia', newsletter: true},
```
kachar, line 9: Let's use emails with suffix `@example.com` so we don't accidentally hit somebody's mailbox :sweat_smile:

db/seed/beneficiary.seed.ts
obozhinov: that's true I'll get them all

kachar (approved):

PR #21, opened 2021-09-20
Added prismadb mock for testing without db instance

igoychev:
Motivation
Currently when executing `yarn test` it fails on city.controler if there is no running database or no cities in it.

Goal
Implement mock for prisma client, which does not require database instance.

Implemented using Prisma Unit test guidelines from here: 
https://www.prisma.io/docs/guides/testing/unit-testing


apps/api/src/prisma/prisma-client-mock.ts
```diff
@@ -0,0 +1,5 @@
+import { PrismaService } from './prisma.service'
+
+const prisma = new PrismaService()
```
kachar, line 3: This file is not actually a mock, maybe we can rename it to `prisma-client.ts`

kachar (approved):
Marvelous upgrade, finally we've got tests working :rocket:

PR #68, opened 2022-01-18
replace campaign summary view with queryraw to fix automatic migrations

quantum-grit:
- upgraded prisma to latest
- replaced campaignSummaryView with prisma.queryRaw
- add migration to drop the campaignSummaryView


apps/api/src/campaign/campaign.service.ts
```diff
@@ -38,15 +56,24 @@ export class CampaignService {
...
     }
+
+    const reachedAmount: Record<string, number> = await this.prisma.$queryRaw`
+      SELECT
+      SUM(d.amount) as reached_amount
+      FROM vaults v
+      JOIN donations d on v.id = d.target_vault_id
+      WHERE d.status = 'succeeded' and v.campaign_id = ${campaign.id}`
+
+    //the query always return 1 record with the value as number or null if no donations where made yet
```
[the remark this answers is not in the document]
igoychev: the frontend handles the null value like this:
const summary = campaign.summary.find(() => true)
const reached = summary ? summary.reachedAmount : 0

imilchev (approved):

apps/api/src/campaign/campaign.service.ts
```diff
@@ -20,11 +20,29 @@ export class CampaignService {
...
 
   async listCampaigns(): Promise<Campaign[]> {
-    return this.prisma.campaign.findMany({
+    const campaigns = await this.prisma.campaign.findMany({
       include: {
-        summary: { select: { reachedAmount: true } },
-      },
+        vaults: {
+          select: {
+            donations: { select: { amount: true} }
```
kachar, line 27: I wonder if we return all donation records to the frontend as well as the other data from the campaign object

Can you post the whole response?

igoychev: yep, indeed vaults and donations get returned to frontend, the TODO on the lines below was for that.
fixed: by setting vaults to [] removes these unnecessary records

kachar (approved):
Smooth transition :level_slider:

Issue #70, opened 2022-01-21
Nestjs warning about annotated imports


kachar:
```
cd api
yarn dev
```

> WARN [DependenciesScanner] In the next major version, Nest will not allow classes annotated with @Injectable(), @Catch(), and @Controller() decorators to appear in the "imports" array of a module.                                                                                                                                             
Please remove "JsonBodyMiddleware" (including forwarded occurrences, if any) from all of the "imports" arrays.                                                                               

> WARN [DependenciesScanner] In the next major version, Nest will not allow classes annotated with @Injectable(), @Catch(), and @Controller() decorators to appear in the "imports" array of a module.                                                                                                                                             
Please remove "RawBodyMiddleware" (including forwarded occurrences, if any) from all of the "imports" arrays.

> Scope [AppModule]

kachar:
Opened discussion in package's repo https://github.com/golevelup/nestjs/discussions/385

quantum-grit:
While upgrading nestjs to 9.1.4 it turned out the stripe middleware imports were not needed - tested and working well without them. The same was suggested also in the package's repo discussion: https://github.com/golevelup/nestjs/discussions/385

PR #72, opened 2022-01-25
Removed middlewares

TodorBelchev:
#70 Removed JsonBodyMiddleware and RawBodyMiddleware from app module imports


kachar:
Thanks for the proposal

I'll take some time to discuss the topic with the maintainers of `webhooks` package

The thread is opened at https://github.com/golevelup/nestjs/discussions/385

I'm not certain about the impacts of this change as the application might break if we just remove the middlewares.

@TodorBelchev  Have you tested the stripe webhooks workflow after the changes have been made?

PR #76, opened 2022-02-04
crud for Country entity (api)

### LINKED ISSUE #75: crud for Country entity (api)
Implement all crud operations for **Country** entity
- [ ] create country
- [ ] get all countries
- [ ] get country by Id
- [ ] update country
- [ ] remove country by Id

dpmarkov:
All tasks from Issue #75 are done.
All tests are passed `yarn nx run-many --target test --all`
Close #75


apps/api/src/country/dto/create-country.dto.ts
```diff
@@ -0,0 +1,22 @@
...
+@Expose()
+export class CreateCountryDto {
+  @ApiProperty()
+  @Expose()
+  @IsString()
+  name: string
+
+  @ApiProperty()
+  @Expose()
+  @IsString()
```
kachar, line 15: We can be more specific here and restrict to iso code `@IsISO31661Alpha2()`

Check https://github.com/typestack/class-validator#validation-decorators

apps/api/src/country/country.controller.ts
```diff
@@ -0,0 +1,40 @@
...
+import { CountryService } from './country.service'
+import { CreateCountryDto } from './dto/create-country.dto'
+import { UpdateCountryDto } from './dto/update-country.dto'
+
+@Controller('country')
+export class CountryController {
+  constructor(private readonly countryService: CountryService) {}
+
+  @Post('create-country')
+  @Public()
```
kachar, line 12: This controller should not be `@Public()` for any action except for `@Get()` requests.

kachar (changes requested):
@dpmarkov Thanks for the effort on extending the codebase.

Please check the notes and let's do another iteration before the merge.

We can't allow public users to delete or create countries in our platform.

We should restrict that only to team members (as in the other controllers)

kachar (approved):

Issue #82, opened 2022-02-06
Create endpoints for CRUD operations with Company entity


TodorBelchev:
Create api resource with Company entity:
 - create
 - edit by id
 - delete by id
 - delete many 
 - read list 
 - read by id

PR #83, opened 2022-02-06
Enhancement/create-company-crud

TodorBelchev:
#82  Create api resource with Company entity:
 - [x] create
- [x] edit by id
- [x] delete by id
- [x] delete many
- [x] read list
- [x] read by id

### DIFF
--- a/apps/api/src/company/company.service.ts
+++ b/apps/api/src/company/company.service.ts
@@ -1,12 +1,12 @@
-import { Injectable, Logger, NotFoundException } from '@nestjs/common';
-import { Prisma } from '@prisma/client';
-import { PrismaService } from '../prisma/prisma.service';
-import { CreateCompanyDto } from './dto/create-company.dto';
-import { UpdateCompanyDto } from './dto/update-company.dto';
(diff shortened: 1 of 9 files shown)


apps/api/src/company/company.service.ts
```diff
@@ -0,0 +1,85 @@
...
+      })
+      if (city == null) {
+        Logger.warn('No city record with ID: ' + cityId)
+        throw new NotFoundException('No city record with ID: ' + cityId)
+      }
+    }
+    return this.prisma.company.create({ data: createCompanyDto })
+  }
+
+  findAll() {
```
kachar, line 27: Method is not `async`

kachar (changes requested):
@TodorBelchev Well done, straight on point

We need to do some tweaks before we merge it so we're in a secure context and we don't leak company data to the public.

Let's do another iteration and update the code

TodorBelchev:
> 

Thank you for the advice. Extracted the interface and replaced it in both account.controller.ts and company.controller.ts.

kachar (approved):
Well done, thanks for the extra iterations, they've upgraded the experience quite a lot.

Issue #85, opened 2022-02-07
[Document] Create endpoints for CRUD operations with Document entity


stilian-nikolaev:
Implement api resource for the Document model with the following CRUD operations:

- [x] create
- [x] edit by id
- [x] delete by id
- [x] delete many
- [x] read list
- [x] read by id

PR #86, opened 2022-02-07
Document entity CRUD

stilian-nikolaev:
[ Issue #85 ](https://github.com/podkrepi-bg/api/issues/85) 
Implement api resource for the Document model with the following CRUD operations:
- [x] create
- [x] edit by id
- [x] delete by id
- [x] delete many
- [x] read list
- [x] read by id


apps/api/src/document/document.controller.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Public } from 'nest-keycloak-connect';
+import { DocumentService } from './document.service';
+import { CreateDocumentDto } from './dto/create-document.dto';
+import { UpdateDocumentDto } from './dto/update-document.dto';
+
+@Controller('document')
+export class DocumentController {
+  constructor(private readonly documentService: DocumentService) { }
+
+  @Public()
```
kachar, line 11: Please make sure that the controller is secured and only authenticated people are able to use the document controller.

Please remove the `@Public()` decorators of all actions

apps/api/src/document/dto/create-document.dto.ts
```diff
@@ -0,0 +1,58 @@
...
+            type: this.type,
+            name: this.name,
+            filename: this.filename,
+            filetype: this.filetype,
+            description: this.description,
+            sourceUrl: this.sourceUrl,
+            owner: {
+                //hardcoded for now
+                connect: {
+                    id: '15661b63-e72f-4c30-8d7d-a75e8d1b69f4'
```
kachar, line 53: Please remove the hardcoded part so we can merge this PR once it's done

apps/api/src/document/dto/create-document.dto.ts
```diff
@@ -0,0 +1,58 @@
+
+import { DocumentType, Prisma } from '@prisma/client'
+import { ApiProperty } from '@nestjs/swagger'
+import { IsNotEmpty, IsObject, IsOptional, IsString, ValidateNested } from 'class-validator';
```
kachar, line 4: Please check the import warnings below

apps/api/src/document/dto/create-document.dto.ts
stilian-nikolaev: @kachar Is it okay to create the owner by passing down the keycloak token like this
From Controller:
`return this.documentService.create(createDocumentDto, user);`
To Service:
`async create(createDocumentDto: CreateDocumentDto, user: KeycloakTokenParsed) {
    return await this.prisma.document.create({ data: createDocumentDto.toEntity(user) })
  }`
To createDocument.toEntity() method in order to connect an existing user or create one according to the data coming from the token?
` owner: {
                connectOrCreate: {
                    where: {
                        email: user.email
                    },
                    create: {
                        firstName: user.given_name,
                        lastName: user.family_name,
                        email: user.email
                    }
                }
            },`

kachar (approved):

apps/api/src/document/dto/create-document.dto.ts
kachar: Yep, this way of handling the owner seems to be good at this point as we don't have a generic approach yet.

Thanks for updating the codebase

PR #87, opened 2022-02-07
Expenses model api

PetyrDzhunov:
image

- GET /api/expenses/list -> lists all the expenses that don't have their deleted boolean to true , a.k.a all expenses that are not deleted.
-GET /api/expenses/id -> returns the data for the special expense with the provided ID
- POST /api/expenses -> create a new expense , by far at the front end I wont be showing the option to connect with documentID and approvedById
- DELETE /api/expenses/id -> soft deleting the passed id (making the deleted boolean to true , yet we still have it in the DB we just dont list it because of our GET /api/expenses/list filter.
- DELETE /api/expenses -> same as the above one , just with the difference that we accept an array with ID's as strings.
- Created a middleware in prisma.service file to make the soft delete possible
- Created seed for vaults which is the parent of the expenses 
- Created seed for expenses , they are all connected to one of the vaults by hardcoding them with the firstly found from the base
- Campaign > Vaults > Expenses , so I created the Vault seed  as an array to the campaign but also they contain array with expenses.
- There is not an update/edit endpoint since logically changing an expense would be illegal?

### DIFF
--- a/apps/api/src/expenses/dto/create-expense.dto.ts
+++ b/apps/api/src/expenses/dto/create-expense.dto.ts
@@ -1,0 +1,1 @@
+export class CreateExpenseDto {}
--- a/apps/api/src/expenses/dto/create-expense.dto.ts
+++ b/apps/api/src/expenses/dto/create-expense.dto.ts
@@ -1,1 +1,65 @@
-export class CreateExpenseDto {}
+import {
+	IsEnum,
+	IsNumber,
+	IsOptional,
+	IsString,
+	MaxLength,
+	MinLength,
+} from 'class-validator'
+import { ApiProperty } from '@nestjs/swagger'
+import { Expose } from 'class-transformer'
+import { Currency, Prisma, ExpenseType, Vault, Person } from '.prisma/client'
+@Expose()
+export class CreateExpenseDto {
+	@ApiProperty()
+	@Expose()
+	type: ExpenseType
+
+	@ApiProperty()
+	@Expose()
+	@IsString()
+	@MinLength(3)
+	@MaxLength(3)
+	@IsEnum(Currency)
+	currency: Currency
+
+	@ApiProperty()
+	@Expose()
+	@IsNumber()
+	amount: number
+
+	@ApiProperty()
+	@Expose()
+	@IsOptional()
+	description: string
+
+	@ApiProperty()
+	@Expose()
+	@IsOptional()
+	vaultId: string
+
+	@ApiProperty()
+	@Expose()
+	@IsOptional()
+	documentId: string
+
+	@ApiProperty()
+	@Expose()
+	@IsOptional()
+	approvedById: string
+
+	@ApiProperty()
+	@Expose()
+	@IsOptional()
+	vault: Vault
+
+	@ApiProperty()
+	@Expose()
+	@IsOptional()
+	approvedBy: Person
+
+	@ApiProperty()
+	@Expose()
+	@IsOptional()
+	document: Document
+}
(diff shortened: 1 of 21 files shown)


apps/api/src/expenses/dto/create-expense.dto.ts
```diff
@@ -0,0 +1,72 @@
...
+//expense->vault ---> one-
+
+// id?: string
+// type: ExpenseType
+// currency?: Currency
+// amount?: number
+// description?: string | null
+// vault: VaultCreateNestedOneWithoutExpensesInput
+// approvedBy?: PersonCreateNestedOneWithoutExpensesInput
+// document?: DocumentCreateNestedOneWithoutExpensesInput
```
quantum-grit, line 72: remove please

apps/api/src/expenses/expenses.service.ts
```diff
@@ -0,0 +1,52 @@
...
+  async findOne(id: string) {
+    try {
+      const expense = await this.prisma.expense.findFirst({ where: { id, deleted: false } })
+      // handle what happens if it finds the id but the deleted is true , what do we return to the client?
+      // may be a message saying the expense has been deleted ?
+      return expense
+    } catch (error) {
+      throw new NotFoundException('No expense found with that id.')
+    }
+  }
```
quantum-grit, line 27: for returning the deleted expenses we can add another function parameter returnDeleted = false and use it in the where clause. This way the user on the frontend can request the deleted expenses if interested in them too.

apps/api/src/prisma/prisma.service.ts
```diff
@@ -4,6 +4,25 @@ import { INestApplication, Injectable, OnModuleInit } from '@nestjs/common'
...
+          if (params.args.data != undefined) {
+            params.args.data['deleted'] = true
+          } else {
+            params.args['data'] = { deleted: true }
+          }
+        }
+      }
+      return next(params)
+    })
+
```
quantum-grit, line 25: nice!

quantum-grit (changes requested):
great job! two small changes and we can merge.

quantum-grit (approved):

apps/api/src/prisma/prisma.service.ts
```diff
@@ -4,6 +4,25 @@ import { INestApplication, Injectable, OnModuleInit } from '@nestjs/common'
 @Injectable()
 export class PrismaService extends PrismaClient implements OnModuleInit {
   async onModuleInit() {
+
+    this.$use(async (params, next) => {
+      if (params.model == 'Expense') {
```
kachar, line 9: This part must be more generic as it is part of the common `PrismaService` but it includes the specific `Expense` entity

apps/api/src/expenses/dto/create-expense.dto.ts
```diff
@@ -0,0 +1,61 @@
...
+  currency: Currency
+
+  @ApiProperty()
+  @Expose()
+  @IsNumber()
+  amount: number
+
+  @ApiProperty()
+  @Expose()
+  @IsString()
```
kachar, line 36: We can add better validation here by saying it's an enum - this will make the enum values the only correct values, not open to any string values as it is right now

apps/api/src/prisma/prisma.service.ts
kachar: We don't want everything to be soft deleted.

There should be a list `string[]` of entities that support soft delete

PetyrDzhunov: I think I understand the point now. So basically I should create an array with strings and manually write a model that needs to be soft deleted inside it , and check if the string[] includes the params.model and if it does then proceed with the soft delete. The question is where to create this array so everybody would find it easily?

PetyrDzhunov: And I think I'm doing something wrong , because I restored the branch so I could make the changes you adviced me to , yet you deleted it , I thought after merging into master I should delete it , can you explain to me what's the proceedure in that case?

PR #90, opened 2022-02-07
Ivan bankaccounts

ivankraev:
# Create Api Endpoints for Bankaccounts model.

- Post request - inserts entity in the database
- Get request - returns all records for the entity from the database
- Get by id - returns unique record selected by id
- Patch request - updates unique entity selected by id
- Delete request - deletes unique record selected by id
- Post request - deletes one or many records selected by id's in array

## Post request
pullpost
## Get request 
pull1
## Get by id 
pullgetbyid
## Patch request 
pullpatch
## Delete request
pulldelete
## Delete Many
pulldeletemany

### DIFF
--- a/apps/api/src/bankaccount/dto/create-bankaccount.dto.ts
+++ b/apps/api/src/bankaccount/dto/create-bankaccount.dto.ts
@@ -1,0 +1,48 @@
+import { ApiProperty } from '@nestjs/swagger'
+import { AccountHolderType, BankAccountStatus } from '@prisma/client'
+import { Expose } from 'class-transformer'
+import { IsEnum, IsOptional, IsString } from 'class-validator'
+import { Withdrawal } from '../../domain/generated/withdrawal/entities/withdrawal.entity'
+@Expose()
+export class CreateBankaccountDto {
+  @ApiProperty({ enum: BankAccountStatus })
+  @Expose()
+  @IsEnum(BankAccountStatus)
+  status: BankAccountStatus
+
+  @ApiProperty()
+  @IsString()
+  @Expose()
+  ibanNumber: string
+
+  @ApiProperty()
+  @Expose()
+  @IsString()
+  accountHolderName: string
+
+  @ApiProperty({ enum: AccountHolderType })
+  @Expose()
+  @IsEnum(AccountHolderType)
+  accountHolderType: AccountHolderType
+
+  @ApiProperty()
+  @Expose()
+  @IsString()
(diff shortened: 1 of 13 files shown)


db/seed/bankaccount.seed.ts
```diff
@@ -0,0 +1,40 @@
+import { PrismaClient } from '@prisma/client'
+const prisma = new PrismaClient()
+
+export async function bankAccountsSeed() {
+  console.log('Cities seed')
+
+  const insert = await prisma.bankAccount.createMany({
+    data: [
+      {
+        accountHolderName: 'Georgi Pavlov',
```
kachar, line 10: You can use random data so we're not using real names by accident

db/seed/index.ts
```diff
@@ -25,6 +26,7 @@ async function main() {
       await coordinatorSeed(),
       await beneficiarySeed(),
       await campaignSeed(),
+      await bankAccountsSeed(),
```
kachar, line 29: Let's make this method singular just like the other sibling functions
```suggestion
      await bankAccountSeed(),
```

migrations/20220201085056_pulling_data/migration.sql
```diff
@@ -0,0 +1 @@
+-- This is an empty migration.
```
kachar, line 1: Looks like we don't need this migration

kachar (changes requested):
@ivankraev Nice iteration! 

Let's do some more work to make sure the bank accounts are not leaked to public and we'll merge it soon

migrations/20220201085056_pulling_data/migration.sql
ivankraev: Thanks for the review ! I will fix the mentioned problems am update the pull request ASAP :)

kachar (approved):

apps/api/src/bankaccount/dto/create-bankaccount.dto.ts
```diff
@@ -0,0 +1,44 @@
...
+  @Expose()
+  @IsString()
+  @IsOptional()
+  bankName: string | null
+
+  @ApiProperty()
+  @Expose()
+  @IsString()
+  @IsOptional()
+  bankIdCode: string | null
```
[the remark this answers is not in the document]
kachar: @ivankraev Please add the validations to the entity in an additional PR as this one will be merged now.

PR #91, opened 2022-02-08
[City] CRUD operations

Yakimov1337:
Screenshot_11


apps/api/src/city/city.controller.ts
```diff
@@ -12,4 +13,28 @@ export class CityController {
...
     return await this.cityService.listCities()
   }
+
+  @Get('list-one/:id')
+  @Public()
+  findOne(@Param('id') id: string) {
+    return this.cityService.listCity(id)
+  }
+
+  @Post('create-city')
```
kachar, line 23: Let's make only get requests `@Public()` but make all others private by removing the `@Public()` annotation

apps/api/src/city/dto/create-city.dto.ts
```diff
@@ -0,0 +1,32 @@
...
+  @Expose()
+  @IsString()
+  name: string
+
+  @ApiProperty()
+  @IsOptional()
+  @Expose()
+  @IsNumber()
+  @IsPositive()
+  @Max(500000)
```
kachar, line 17: Please use the validator `@IsPostalCode(locale?: string)` 

Not sure `500000` is the actual max for the postcode, let's just use the postal code validator for now

Check the docs at https://github.com/typestack/class-validator#validation-decorators

apps/api/src/city/city.controller.ts
```diff
@@ -12,4 +13,28 @@ export class CityController {
   async getData() {
     return await this.cityService.listCities()
   }
+
+  @Get('list-one/:id')
```
kachar, line 17: Please remove the `list-one/` as it not by the RESTful standard and makes the URL `/city/list-one/12345` instead of `/city/12345`

```suggestion
  @Get(':id')
```

apps/api/src/city/city.controller.ts
```diff
@@ -12,4 +13,28 @@ export class CityController {
...
+    return this.cityService.listCity(id)
+  }
+
+  @Post('create-city')
+  @Public()
+  async create(@Body() CreateCityDto: CreateCityDto) {
+    return await this.cityService.createCity(CreateCityDto)
+  }
+
+  @Patch('update-one/:id')
```
kachar, line 29: Please remove the `update-one/` from the path

```suggestion
  @Patch(':id')
```

apps/api/src/city/city.controller.ts
```diff
@@ -12,4 +13,28 @@ export class CityController {
...
+  @Public()
+  update(@Param('id') id: string, @Body() updateCityDto: UpdateCityDto) {
+    return this.cityService.updateCity(id, updateCityDto)
+  }
+
+  @Delete('delete-one/:id')
```
kachar, line 35:
```suggestion
  @Delete(':id')
```

kachar (changes requested):
@Yakimov1337 Nice work, well done

Let's do one more iteration and make sure the content creation/deletion is secure before we merge the PR.

Please also check the other remarks

PR #92, opened 2022-02-08
[Bootcamp] InfoRequest model

dplamenov:
Create CRUD for **InfoRequest** model.
These are new paths: 
info-request

### DIFF
--- a/apps/api/src/info-request/dto/create-info-request.dto.ts
+++ b/apps/api/src/info-request/dto/create-info-request.dto.ts
@@ -1,0 +1,1 @@
+export class CreateInfoRequestDto {}
--- a/apps/api/src/info-request/dto/create-info-request.dto.ts
+++ b/apps/api/src/info-request/dto/create-info-request.dto.ts
@@ -1,11 +1,12 @@
 import { ApiProperty } from '@nestjs/swagger'
 import { Expose } from 'class-transformer'
-import { IsString } from 'class-validator'
+import { IsString, IsUUID } from 'class-validator'
 
 export class CreateInfoRequestDto {
     @ApiProperty()
     @Expose()
     @IsString()
+    @IsUUID()
     personId: string;
     @ApiProperty()
     @Expose()
--- a/apps/api/src/info-request/dto/create-info-request.dto.ts
+++ b/apps/api/src/info-request/dto/create-info-request.dto.ts
@@ -1,1 +1,14 @@
-export class CreateInfoRequestDto {}
+import { ApiProperty } from '@nestjs/swagger'
+import { Expose } from 'class-transformer'
+import { IsString } from 'class-validator'
+
+export class CreateInfoRequestDto {
+    @ApiProperty()
+    @Expose()
+    @IsString()
+    personId: string;
+    @ApiProperty()
+    @Expose()
+    @IsString()
+    message: string;
+}
--- a/apps/api/src/info-request/info-request.controller.ts
+++ b/apps/api/src/info-request/info-request.controller.ts
@@ -1,8 +1,8 @@
 import { Controller, Get, Post, Body, Patch, Param, Delete } from '@nestjs/common';
 import { InfoRequestService } from './info-request.service';
-import { CreateInfoRequestDto } from './dto/create-info-request.dto';
 import { UpdateInfoRequestDto } from './dto/update-info-request.dto';
 import { Public } from 'nest-keycloak-connect';
+import { CreateInquiryDto } from '../support/dto/create-inquiry.dto';
 
 @Controller('info-request')
 export class InfoRequestController {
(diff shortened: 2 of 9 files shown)


apps/api/src/info-request/info-request.service.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@Injectable()
+export class InfoRequestService {
+  constructor(public prisma: PrismaService){}
+
+  create(createInfoRequestDto: CreateInfoRequestDto) {
+    return this.prisma.infoRequest.create({data: createInfoRequestDto});
+  }
+
+  findAll() {
```
kachar, line 14: @dplamenov  At the moment partially this service has been done in `SupportService`

https://github.com/podkrepi-bg/api/blob/b3ec755bb400e55715400080e3c6306d4c306a5f/apps/api/src/support/support.service.ts#L40-L56

It's kind of duplication as the creation of info request would be expected to do the same thing and at the moment it does two separate things.

Here are some ideas how to go further:

1. Remove the creation methods from `InfoRequestService`
2. Move the logic from `SupportService` to `InfoRequest` service and make them dependent
3. Something else that you might think of


Let's discuss the further steps to refactor this scenario so the code is located only on one place

apps/api/src/info-request/info-request.controller.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  @Get(':id')
+  @Public()
+  findOne(@Param('id') id: string) {
+    return this.infoRequestService.findOne(id);
+  }
+
+  @Patch(':id')
+  @Public()
```
kachar, line 30: Let's make all editorial endpoints actions private, so noone is able to edit a info request without permissions

```suggestion
```

apps/api/src/info-request/info-request.controller.ts
```diff
@@ -0,0 +1,40 @@
...
+  update(@Param('id') id: string, @Body() updateInfoRequestDto: UpdateInfoRequestDto) {
+    return this.infoRequestService.update(id, updateInfoRequestDto);
+  }
+
+  @Delete(':id')
+  @Public()
```
kachar, line 36:
```suggestion
```

apps/api/src/info-request/info-request.controller.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  @Get('list')
+  @Public()
+  findAll() {
+    return this.infoRequestService.findAll();
+  }
+
+  @Get(':id')
+  @Public()
```
kachar, line 24:
```suggestion
```

apps/api/src/info-request/dto/create-info-request.dto.ts
```diff
@@ -0,0 +1,14 @@
+import { ApiProperty } from '@nestjs/swagger'
+import { Expose } from 'class-transformer'
+import { IsString } from 'class-validator'
+
+export class CreateInfoRequestDto {
```
kachar, line 5: This class actually copies the functionality of `CreateInquiryDto`

https://github.com/podkrepi-bg/api/blob/c4355b6f8221dc7a0e80ab235c55e142f3493174/apps/api/src/support/dto/create-inquiry.dto.ts#L8

Let's leave them co-exsist for now and will check later on for refactoring

kachar (changes requested):
@dpetrov Let's discuss how to go further with the refactoring of this PR as it copies already working functionallity

apps/api/src/info-request/info-request.service.ts
dplamenov: Can I delete the InfoRequestService and use SupportService at InfoRequestController?

dplamenov: I committed changes but I am not sure is already ok.

kachar (approved):

apps/api/src/info-request/info-request.service.ts
```diff
@@ -0,0 +1,30 @@
+import { Injectable } from '@nestjs/common';
+import { PrismaService } from '../prisma/prisma.service';
+import { CreateInquiryDto } from '../support/dto/create-inquiry.dto';
+import { SupportService } from '../support/support.service';
+import { UpdateInfoRequestDto } from './dto/update-info-request.dto';
+
+@Injectable()
+export class InfoRequestService {
+  constructor(public prisma: PrismaService, public supportService: SupportService){}
```
kachar, line 9:
```suggestion
  constructor(private prisma: PrismaService, private supportService: SupportService){}
```

apps/api/src/info-request/info-request.controller.ts
```diff
@@ -0,0 +1,37 @@
...
+  constructor(private readonly infoRequestService: InfoRequestService) {}
+
+  @Post()
+  @Public()
+  create(@Body() createInfoRequestDto: CreateInquiryDto) {
+    return this.infoRequestService.create(createInfoRequestDto);
+  }
+
+  @Get('list')
+  @Public()
```
kachar, line 18:
```suggestion
```

PR #104, opened 2022-02-21
[City entity] PostalCode and service fix

Yakimov1337:
1. PostalCode changed from Number to String
2. Updated city service (findMany includes countryCode)


kachar (approved):
@Yakimov1337 Looks good, thanks for the update

I don't get why you've made so many small changes (reorder) to the schema, but that's fine. I've patched the conflicts and now it looks good. You needed to run `yarn generate-schema` so your changes are applied fully to the codebase, not only to the schema file.

Thanks for your contribution :rocket:

Issue #107, opened 2022-02-22
[Vault] Implement CRUD operations for Vault entity


stilian-nikolaev:
Implement api resource for the Vault model with the following CRUD operations:

- [x]  create
- [x]  edit by id
- [x]  delete by id
- [x]  delete many
 - [x] read list
- [x]  read by id

PR #110, opened 2022-02-24
Vault entity crud

stilian-nikolaev:
Closing [#107 ](https://github.com/podkrepi-bg/api/issues/107)

providing the following crud functionality:

- create
- edit by id
- delete by id
- delete many
- read list
 -read by id

added a name field to the prisma schema
a couple conflicts occurred in the campaign service and createDto due to the changes in the Vault model
I have added a temporary example name to dismiss the errors

I have also added name field to the vault seed through faker

through updateDto only name can be updated

I am not sure yet how to deal with the other relations of the model
I am open to suggestions to improve this pull request :)

### DIFF
--- a/apps/api/src/vault/dto/create-vault.dto.ts
+++ b/apps/api/src/vault/dto/create-vault.dto.ts
@@ -1,0 +1,1 @@
+export class CreateVaultDto {}
--- a/apps/api/src/vault/dto/create-vault.dto.ts
+++ b/apps/api/src/vault/dto/create-vault.dto.ts
@@ -1,6 +1,5 @@
 import { ApiProperty } from '@nestjs/swagger'
 import {
-  Campaign,
   Currency,
   Donation,
   Expense,
--- a/apps/api/src/vault/dto/create-vault.dto.ts
+++ b/apps/api/src/vault/dto/create-vault.dto.ts
@@ -10,7 +10,7 @@ import {
   Withdrawal,
 } from '@prisma/client'
 import { Expose } from 'class-transformer'
-import { IsDate, IsNumber, IsOptional, IsPositive, IsString, IsUUID } from 'class-validator'
+import { IsOptional, IsString, IsUUID } from 'class-validator'
 
 @Expose()
 export class CreateVaultDto {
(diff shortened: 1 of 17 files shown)


apps/api/src/vault/vault.controller.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  @Post()
+  create(@AuthenticatedUser() user: KeycloakTokenParsed,
+    @Body() createVaultDto: CreateVaultDto) {
+    if (!user) {
+      throw new UnauthorizedException()
+    }
+
+    return this.vaultService.create(createVaultDto)
+  }
```
quantum-grit, line 21: For this and all methods changing the vault: Only the coordinator can create or change Vaults. So, here we have to check the user.id with the coordinatorID of the campaign.

apps/api/src/vault/dto/create-vault.dto.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  @Expose()
+  @ApiProperty()
+  @IsOptional()
+  withdraws?: Withdrawal[]
+
+  @Expose()
+  @ApiProperty()
+  @IsOptional()
+  recurringDonations?: RecurringDonation[]
```
quantum-grit, line 59: When the vault is created these will always be empty, so please remove them from the create-vault.dto. They will happen through the operations of Donation and making Expenses/Withdrawal later.

apps/api/src/vault/vault.service.ts
```diff
@@ -0,0 +1,88 @@
...
+          id: {
+            in: idsToDelete,
+          },
+        },
+      })
+    } catch (err) {
+      const msg = `Delete failed. No Vault found with given ID`
+
+      Logger.warn(msg)
+      throw new NotFoundException(msg)
```
quantum-grit, line 85: here the error may not be only that the ID is not found, so please rethrow the original error. The code should be like this or better:
 } catch (err) {
      const msg = `Error while deleting Vault with id: ` + idsToDelete + " Exception was: " + err.message

      Logger.warn(msg)
      throw err

quantum-grit (changes requested):
good job, please check the comments

quantum-grit (approved):

PR #112, opened 2022-02-25
Added delete endpoint for beneficiary entity

PepiPetrov:
Added delete endpoint for beneficiary entity


.husky/workspace.json
```diff
@@ -0,0 +1,82 @@
...
+          }
+        }
+      }
+    }
+  },
+  "cli": {
+    "defaultCollection": "@nrwl/nest"
+  },
+  "defaultProject": "api"
+}
```
kachar, line 82: Please clean up the `.husky` folder as it contains extra files

.husky/pre-commit
```diff
@@ -1,4 +0,0 @@
-#!/bin/sh
```
kachar, line 1: Please restore the `.husky/pre-commit` and `.husky/.gitignore` files as they are part of the lint process on commit.

apps/api/src/city/dto/createBootcamp.dto.ts
```diff
@@ -0,0 +1,5 @@
+export class CityCreateDto {
```
kachar, line 1: I don't think we need entity with name `createBootcamp` in the repo.

Is there a reason for it ?

PepiPetrov: This is because i merged the code from master. I will delete it.

apps/api/src/person/dto/create-person.dto.ts
```diff
@@ -0,0 +1,20 @@
+
+
+
+
+
+
+export class CreatePersonDto {
```
kachar, line 7: Please run `yarn format` in the repo and commit the resulting files as there are multiple files with broken indentation and code style

apps/api/src/city/dto/createBootcamp.dto.ts
PepiPetrov: Done.

migrations/20210912131923_init/migration.sql
```diff
@@ -137,7 +137,7 @@ ALTER TABLE "beneficiaries" ADD CONSTRAINT "beneficiaries_personId_fkey" FOREIGN
 ALTER TABLE "beneficiaries" ADD CONSTRAINT "beneficiaries_coordinatorId_fkey" FOREIGN KEY ("coordinatorId") REFERENCES "coordinators"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
 
 -- AddForeignKey
-ALTER TABLE "campaigns" ADD CONSTRAINT "campaigns_beneficiary_id_fkey" FOREIGN KEY ("beneficiary_id") REFERENCES "beneficiaries"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
```
kachar, line 140: Please restore the initial version of the migration as it was before. If we need to change something in the database it needs to happen in the following way:

1. Update of `schema.prisma
2. Generate client and entities `yarn generate-schema`
3. Generate new migration with `yarn prisma migrate dev`

migrations/20220218091359_fixed_campaign_type_model/migration.sql
```diff
@@ -0,0 +1,74 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `status` on the `expenses` table. All the data in the column will be lost.
+  - You are about to drop the column `address` on the `people` table. All the data in the column will be lost.
```
kachar, line 5: Why do we drop the column `address` on the `people` table?

PepiPetrov: I'll check it.

kachar (commented):
Please remove `stripe.exe` from the git repo as it's not needed, it has different way of distribution

migrations/20220221155611_fixed_on_delete_for_other_related_tables/migration.sql
```diff
@@ -0,0 +1,17 @@
+-- DropForeignKey
```
kachar, line 1: If there's no update of `schema.prisma` then we should not have any new data migrations

migrations/20220218091359_fixed_campaign_type_model/migration.sql
PepiPetrov: Done.

migrations/20210912131923_init/migration.sql
PepiPetrov: Done.

migrations/20220221155611_fixed_on_delete_for_other_related_tables/migration.sql
PepiPetrov: Done.

.husky/pre-commit
PepiPetrov: Done.

apps/api/src/beneficiary/beneficiary.controller.ts
```diff
@@ -1,26 +1,58 @@
...
+  })
   async create(@Body() createDto: CreateBeneficiaryDto) {
     return await this.beneficiaryService.createBeneficiary(createDto)
   }
 
   @Get('list')
+  // @Roles({
+  //   roles: [RealmViewSupporters.role, ViewSupporters.role],
+  //   mode: RoleMatchingMode.ANY,
+  // })
```
kachar, line 26: If we need this code to effective please remove the comment, if we don;t need this code please remove the code.

Not sure what should be the exact action here, maybe de-comment

migrations/20220221155426_fixed_on_delete/migration.sql
```diff
@@ -0,0 +1,23 @@
+-- DropForeignKey
+ALTER TABLE "transfers" DROP CONSTRAINT "transfers_source_campaign_id_fkey";
```
kachar, line 2: This migration has no code update in `prisma.schema`

apps/api/src/beneficiary/beneficiary.controller.ts
PepiPetrov: This is for authentication.

PepiPetrov: Fixed.

kachar (approved):
Well done @PepiPetrov 

Thanks for developing this part of the system

Let's split the entities into separate branches and PRs next time for independent review and integration. It will be easier to handle also.

PR #117, opened 2022-03-02
Removed delete many beneficiary dto

PepiPetrov:
Removed delete many beneficiary dto.


.husky/.prettierignore
```diff
@@ -0,0 +1,5 @@
+# Add files here to ignore them from prettier formatting
```
kachar, line 1: @PepiPetrov Looks like the .husky folder got polluted again with some other files

Please clean it up so there is no diff in the PR for this folder

kachar (changes requested):
@PepiPetrov Looks like the .husky folder got polluted again with some other files

Please clean it up so there is no diff in the PR for this folder

Maybe you started the branch from different point of history instead of latest master

PR #123, opened 2022-03-07
Create campaign s3 upload

imilchev:
This PR implements file uploads for campaigns.

- A new database table has been added to store campaign files
- Extended campaign retrieval to return a list of files related to it
- S3 integration has been implemented to store the uploaded files
  - Possibility to upload multiple files at once
  - Files' mimetype is preserved in S3 while uploading
  - Files are stored using the PK from the db in the S3 bucket to ensure uniqueness
  - When downloading files, they are being returned with their original names

### DIFF
--- a/apps/api/src/campaign-file/campaign-file.controller.ts
+++ b/apps/api/src/campaign-file/campaign-file.controller.ts
@@ -1,34 +1,54 @@
-import { Controller, Get, Post, Body, Patch, Param, Delete } from '@nestjs/common'
+import {
+  Controller,
+  Get,
+  Post,
+  Body,
+  Response,
+  Param,
+  Delete,
+  StreamableFile,
+} from '@nestjs/common'
 import { CampaignFileService } from './campaign-file.service'
-import { CreateCampaignFileDto } from './dto/create-campaign-file.dto'
 import { Public } from 'nest-keycloak-connect'
-import { Logger, LogLevel, VersioningType } from '@nestjs/common'
+import { Logger, UseInterceptors, UploadedFiles } from '@nestjs/common'
+import { FilesInterceptor } from '@nestjs/platform-express'
+import { Multer } from 'multer'
 
 @Controller('campaign-file')
 export class CampaignFileController {
   constructor(private readonly campaignFileService: CampaignFileService) {}
 
-  @Post()
+  @Post(':campaign_id')
   @Public()
-  create(@Body() createCampaignFileDto: CreateCampaignFileDto) {
-    console.log('hmm')
-    Logger.log('hmm')
-    return this.campaignFileService.create(createCampaignFileDto)
-  }
-
-  @Get(':campaign_id')
-  @Public()
-  findAll(@Param('campaign_id') campaignId: string) {
-    Logger.log('hmm')
-    return this.campaignFileService.findAll(campaignId)
+  @UseInterceptors(FilesInterceptor('file'))
+  async create(
+    @Param('campaign_id') campaignId: string,
+    @UploadedFiles() files: Express.Multer.File[],
+  ) {
+    return await Promise.all(
+      files.map((x) =>
+        this.campaignFileService.create(campaignId, x.originalname, x.mimetype, x.buffer),
+      ),
+    )
   }
 
   @Get(':id')
-  findOne(@Param('id') id: string) {
-    return this.campaignFileService.findOne(id)
+  @Public()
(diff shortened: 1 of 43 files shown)


apps/api/src/campaign-file/campaign-file.controller.ts
```diff
@@ -0,0 +1,55 @@
...
+import { Logger, UseInterceptors, UploadedFiles } from '@nestjs/common'
+import { FilesInterceptor } from '@nestjs/platform-express'
+import { Multer } from 'multer'
+
+@Controller('campaign-file')
+export class CampaignFileController {
+  constructor(private readonly campaignFileService: CampaignFileService) {}
+
+  @Post(':campaign_id')
+  @Public()
```
imilchev, line 22: not sure what access we want to configure for this. I saw that creating a campaign is also public so I kept this the same.

kachar (approved):
Awesome integration of S3, thanks @imilchev 

One addition that might do is to update the kubernetes manifests with the required env vars, but that can be done in separate PR

apps/api/src/campaign/campaign.service.ts
```diff
@@ -73,6 +73,7 @@ export class CampaignService {
             person: { select: { id: true, firstName: true, lastName: true } },
           },
         },
+        campaignFiles: true,
```
igoychev, line 76: may be the campaign files are not needed to be returned each time, especially if the list is long. The suggestion is to load the file list in separate query where paging could be implemented too.

apps/api/src/s3/s3.service.ts
```diff
@@ -0,0 +1,56 @@
...
+    return await this.s3
+      .upload({
+        Bucket: this.bucketName,
+        Body: stream,
+        Key: key,
+        ContentType: mimetype,
+      })
+      .promise()
+      .then((x) => x.Key)
+  }
```
igoychev, line 39: while uploading the file with guid as name, let's add some metadata fields to store the real filename and campaign Id/name, username of who uploaded it for easier exploration and debuging. Example:
Metadata: {origFilename: 'document.pdf', uploader: 'username' , campaignId: 'guid', campaignTitle: "campaign.title"},

schema.prisma
```diff
@@ -177,10 +177,20 @@ model Campaign {
...
 }
 
+model CampaignFile {
+  id         String   @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
+  filename   String   @db.VarChar(200)
+  campaignId String   @map("campaign_id") @db.Uuid
+  campaign   Campaign @relation(fields: [campaignId], references: [id])
+
+  @@map("campaign_files")
+}
```
igoychev, line 192: wondering if we need an uploadedBy: Person relation if not only the coordinator can upload a file.
do we also need the createdAt field or we will take that from S3?

apps/api/src/campaign-file/campaign-file.controller.ts
igoychev: I think it should be allowed only for campaign coordinator, beneficiery or campaign admin roles

igoychev (changes requested):
Nice and neat - well done! 
I've added suggestions regarding the permissions and additional metadata when saving the files.

apps/api/src/campaign/campaign.service.ts
imilchev: fair point. I will add an endpoint to get the campaign files for a campaign

imilchev: This is actually the "campaign details" query. So this is only returned when you get a single campaign. Listing the campaigns does not include the campaign files.

apps/api/src/s3/s3.service.ts
imilchev: I added the uploaded by, campaign and filename in S3. I think the IDs should be enough for debugging purposes since the names could potentially change in the DB but the IDs should be immutable

schema.prisma
imilchev: I added that and is now retrieved from the auth token. Meaning that we can restrict the access to that endpoint based on roles in the controller but the business logic will always work.

apps/api/src/campaign/campaign.service.ts
igoychev: ok, let's leave it as is and we can optimize later.

igoychev (approved):

PR #133, opened 2022-03-11
Fixed campaign types list endpoint to be public

PepiPetrov:
Fixed campaign types list endpoint to be public


kachar (approved):
Nice catch

PR #143, opened 2022-03-17
Adding fields (file type and role) to campaign file entity.

PepiPetrov:
Adding fields (file type and role) to campaign file entity.


apps/api/src/campaign-file/campaign-file.service.ts
```diff
@@ -3,18 +3,16 @@ import { PrismaService } from '../prisma/prisma.service'
 import { S3Service } from '../s3/s3.service'
 import { CreateCampaignFileDto } from './dto/create-campaign-file.dto'
 import { Readable } from 'stream'
-import { PersonService } from '../person/person.service'
+// import { PersonService } from '../person/person.service'
```
imilchev, line 6: either delete this line or uncomment it

apps/api/src/campaign-file/campaign-file.controller.ts
```diff
@@ -37,7 +37,14 @@ export class CampaignFileController {
     }
     return await Promise.all(
       files.map((x) =>
-        this.campaignFileService.create(campaignId, x.originalname, x.mimetype, person, x.buffer),
+        this.campaignFileService.create(
+          campaignId,
+          'background',
```
imilchev, line 42: I think we should use an enumeration for this. One option can be background photo, another could be document or invoice. @kachar what do you think?

Another issue is that you are hardcoding the `'background'` now, when it should be coming from the front-end instead. When the user uploads a file he/she should decide what the role of that file is

migrations/20220213130743_/migration.sql
```diff
@@ -1,8 +0,0 @@
-/*
-  Warnings:
-
-  - A unique constraint covering the columns `[person_id]` on the table `coordinators` will be added. If there are existing duplicate values, this will fail.
-
-*/
--- CreateIndex
-CREATE UNIQUE INDEX "coordinators_person_id_key" ON "coordinators"("person_id");
```
imilchev, line 8: why is this migration deleted?

imilchev (changes requested):

apps/api/src/campaign-file/campaign-file.controller.ts
kachar: I do agree on both points :+1:

schema.prisma
```diff
@@ -184,12 +184,15 @@ model Campaign {
...
-  id            String   @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
-  filename      String   @db.VarChar(200)
-  campaignId    String   @map("campaign_id") @db.Uuid
-  campaign      Campaign @relation(fields: [campaignId], references: [id])
-  uploadedById  String   @map("uploaded_by_id") @db.Uuid
-  uploadedBy    Person   @relation(fields: [uploadedById], references: [id])
+  id           String   @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
+  filename     String   @db.VarChar(200)
+  role         String   @db.VarChar(200)
+  type         String   @db.VarChar(200)
```
kachar, line 190: We better have two enum records for each of the `role` and `type` dimensions

```prisma
enum FileRole {
  ...
  @@map("file_role")
}

Comment (imilchev) [reply schema.prisma:190]: I agree

File: apps/api/src/campaign-file/campaign-file.controller.ts (hunk 2 above)
Comment (kachar) [reply apps/api/src/campaign-file/campaign-file.controller.ts:42]: @PepiPetrov Please use the new enum in this location so the hardcoded string is replaced

File: apps/api/src/campaign-file/campaign-file.service.ts [hunk 5]
```diff
@@ -3,18 +3,15 @@ import { PrismaService } from '../prisma/prisma.service'
...
 export class CampaignFileService {
-  constructor(
-    private prisma: PrismaService,
-    private s3: S3Service,
-  ) {}
+  constructor(private prisma: PrismaService, private s3: S3Service) {}
 
   async create(
     campaignId: string,
+    role: string,
```
kachar, line 14: Please use the enum as type hint for the `role` arg

apps/api/src/campaign-file/dto/create-campaign-file.dto.ts
```diff
@@ -2,4 +2,6 @@ export class CreateCampaignFileDto {
   filename: string
   campaignId: string
   uploadedById: string
+  role: string
```
kachar, line 5: Please use the enum here as well

migrations/20220201085056_pulling_data/migration.sql
```diff
@@ -1 +0,0 @@
--- This is an empty migration.
```
kachar, line 1: Please undo the deletion of this migration, even though it's empty its part of the migration history which needs it

schema.prisma
```diff
@@ -618,3 +621,20 @@ enum BankAccountStatus {
...
 }
+
+enum CampaignFileRole {
+  background
+  coordinator
+  campaignPhoto
+  invoice
+  document
+
+  @@map("role")
```
kachar, line 632: This should be `campaign_file_role` as it will be very very confusing when the column name is `role` and column type is `role`

Should be 

```suggestion
  @@map("campaign_file_role")
```

migrations/20220318140415_added_enums_for_campaign_file_role_and_type/migration.sql
```diff
@@ -0,0 +1,16 @@
...
+*/
+-- CreateEnum
+CREATE TYPE "role" AS ENUM ('background', 'coordinator', 'campaignPhoto', 'invoice', 'document');
+
+-- CreateEnum
+CREATE TYPE "CampaignFileType" AS ENUM ('png', 'jpeg', 'pdf', 'docx');
+
+-- AlterTable
+ALTER TABLE "campaign_files" DROP COLUMN "role",
+ADD COLUMN     "role" "role" NOT NULL;
```
kachar, line 16:
```suggestion
CREATE TYPE "campaign_file_role" AS ENUM ('background', 'coordinator', 'campaignPhoto', 'invoice', 'document');

-- CreateEnum
CREATE TYPE "CampaignFileType" AS ENUM ('png', 'jpeg', 'pdf', 'docx');

-- AlterTable
ALTER TABLE "campaign_files" DROP COLUMN "role",
ADD COLUMN     "role" "campaign_file_role" NOT NULL;
```

package.json
```diff
@@ -80,7 +80,6 @@
     "jest-extended": "^1.2.0",
     "jest-mock-extended": "^2.0.4",
     "lint-staged": "11.1.2",
-    "nx": "12.9.0",
```
kachar, line 83: Why `nx` is removed from the `devDependencies`? 

I think we better have it in as many scripts depend on it's binaries.

Please roll it back

schema.prisma
```diff
@@ -184,12 +184,15 @@ model Campaign {
...
-  id            String   @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
-  filename      String   @db.VarChar(200)
-  campaignId    String   @map("campaign_id") @db.Uuid
-  campaign      Campaign @relation(fields: [campaignId], references: [id])
-  uploadedById  String   @map("uploaded_by_id") @db.Uuid
-  uploadedBy    Person   @relation(fields: [uploadedById], references: [id])
+  id           String           @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
+  filename     String           @db.VarChar(200)
+  role         CampaignFileRole
+  type         String           @db.VarChar(200)
```
kachar, line 190: This type is still string, maybe we should use `CampaignFileType` enum instead.

If `CampaignFileType` will not be used we better delete it

kachar (approved):
@PepiPetrov Thanks for the contribution

I had to alter some of the code prior the merge as it had extra files and steps.

I've also removed the limitation for the file type as it was way too limiting and we better roll with the pure content type instead of the extension of the file.

PR #150, opened 2022-03-22
Bootcamp first entity  - neli radkova

nradkova:


apps/api/src/bootcampNeli/bootcampNeli.controller.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Post('create')
+  @Public()
+  create(@Body() creatBootcampNeliDto: CreateBootcampNeliDto) {
+    return this.bootcampNeliService.createBootcampNeli(creatBootcampNeliDto)
+  }
+
+  @Put(':id')
+  @Public()
```
kachar, line 32: In real scenarios we should not expose the API with `@Public()` as it means everybody can change your data

schema.prisma
```diff
@@ -29,6 +29,14 @@ datasource db {
   url      = env("DATABASE_URL")
 }
 
+model BootcampNeli {
+  id                 String              @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
+  firstName          String              @map("first_name") @db.VarChar(100)
+  lastName           String              @map("last_name") @db.VarChar(100)
+  email              String              @unique @db.Citext
+   @@map("bootcamp_Neli")
```
kachar, line 37: The db tables and columns should always be lowercase

I understand here it's a bit different because of the personal name, but that's the general rule

kachar (approved):
Well done @nradkova 

That's all what's expected from the demo task, I feel you're ready for a real task now

PR #151, opened 2022-03-22
Tests imilchev

borislavstoychev:


imilchev:
I fixed it on master so it should work the next time the tests are run

PR #155, opened 2022-03-23
Added mimetype specification

PepiPetrov:
Added mimetype specification becase header for specifying mimetype was set to 'application/json'.


imilchev:
This is not needed as you should already have the mimetype in the database. I implement a fix for the issue in #156

Issue #154, opened 2022-03-23
Specifying mimetype


PepiPetrov:
Added specifcation for mimetype because the mimetype sent was 'application/json'.

PR #156, opened 2022-03-23
Set correct content type when retrieving a campaign file

imilchev:
- make mimetype a mandatory column in the database
- set correct content type when retrieving a campaign file

@kachar I saw that you added the mimetype column in the db but I was not sure why you made it optional. Since I changed this now I am adding you as a reviewer to check if I missed anything


imilchev:
fixes #154

kachar (approved):
@imilchev Thanks for the PR

I was not aware if all files need to have their content type set, I see it's a good thing so good call on making it mandatory.

PR #158, opened 2022-03-23
Added extra roles to enum CampaignFileRole

PepiPetrov:
Added extra roles to enum CampaignFileRole. The new roles are anyCampaignFile - this is a role for the circular photo above the campaign info, and campaignListPhoto - this is a role for the photo for the list of campaigns.


schema.prisma
```diff
@@ -628,6 +628,8 @@ enum CampaignFileRole {
   campaignPhoto
   invoice
   document
+  anyCampaignPicture
```
kachar, line 631: In order to be consice let's rename `anyCampaignPicture` to `profilePhoto`
```suggestion
  profilePhoto
```

All other enum values are photo, not picture

The main function is to be profile photo, not any photo

kachar (approved):

kachar (approved):

Issue #146, opened 2022-03-18
Add category to campaign type entity


kachar:
In order to group the campaigns by their category belonging we need to add a property `category` to each `CampaignType` entity

Let's define that as new enum `CampaginTypeCategory` containing the following items:

- `medical`
- `charity`
- `disasters`
- `education`
- `events`
- `environment`
- `sport`
- `art`
- `others`

It should be added to:

- prisma schema file
- schema generators
- seeds

PR #160, opened 2022-03-24
[CampaignType] Add category column for campaign type entity

Yakimov1337:
- [x] Added category column for campaign type based on #146
Screenshot_38
6

### DIFF
--- a/schema.prisma
+++ b/schema.prisma
@@ -141,7 +141,7 @@ model Beneficiary {
 
 model CampaignType {
   id          String               @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
-  category    CampaignTypeCategory
+  category    CampaignTypeCategory @default(others)
   name        String               @db.VarChar(50)
   slug        String               @unique @db.VarChar(50)
   description String?              @db.VarChar(200)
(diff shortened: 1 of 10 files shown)


schema.prisma
```diff
@@ -140,13 +140,14 @@ model Beneficiary {
...
 model CampaignType {
-  id          String         @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
-  name        String         @db.VarChar(50)
-  slug        String         @unique @db.VarChar(50)
-  description String?        @db.VarChar(200)
-  parentId    String?        @map("parent_id") @db.Uuid
-  parent      CampaignType?  @relation("ParentCategory", fields: [parentId], references: [id])
-  children    CampaignType[] @relation("ParentCategory")
+  id          String               @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
+  category    CampaignTypeCategory
```
kachar, line 144: @Yakimov1337 I'll add a default value for the column otherwise the db will throw error for the existing records when we run the migration

kachar: There's a warning in the migration file:

> Added the required column `category` to the `campaign_types` table without a default value. This is not possible if the table is not empty.

kachar (approved):
Flawless upgrade @Yakimov1337 

Well done!

kachar (approved):

PR #161, opened 2022-03-24
Campaign file upload

dplamenov:
Store file role in database

### DIFF
--- a/apps/api/src/campaign-file/campaign-file.controller.ts
+++ b/apps/api/src/campaign-file/campaign-file.controller.ts
@@ -16,7 +16,6 @@ import { Public, AuthenticatedUser } from 'nest-keycloak-connect'
 import { UseInterceptors, UploadedFiles } from '@nestjs/common'
 import { FilesInterceptor } from '@nestjs/platform-express'
 import { PersonService } from '../person/person.service'
-import { CampaignFileRole } from '@prisma/client'
 
 @Controller('campaign-file')
 export class CampaignFileController {
--- a/apps/api/src/campaign-file/campaign-file.controller.ts
+++ b/apps/api/src/campaign-file/campaign-file.controller.ts
@@ -9,6 +9,7 @@ import {
   StreamableFile,
   NotFoundException,
   Logger,
+  Body,
 } from '@nestjs/common'
 import { CampaignFileService } from './campaign-file.service'
 import { Public, AuthenticatedUser } from 'nest-keycloak-connect'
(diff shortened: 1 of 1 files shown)


apps/api/src/campaign-file/campaign-file.controller.ts
```diff
@@ -28,9 +28,11 @@ export class CampaignFileController {
   @UseInterceptors(FilesInterceptor('file'))
   async create(
     @Param('campaign_id') campaignId: string,
+    @Body() body: { filesRole: string },
```
kachar, line 31: @dplamenov Can you please extract this var in dedicated DTO object

In such way you can filter and validate the input prior hitting the controller

dplamenov: @kachar, I don't know how to validate `filesRole` because it is a string. Can you provide more details? Please.

dplamenov:
I created new DTO - `FilesRoleDto` because `CreateCampaignFileDto` is used in service.

kachar (approved):
Awesome work @dplamenov

apps/api/src/campaign-file/dto/files-role.dto.ts
```diff
@@ -0,0 +1,11 @@
+import { ApiProperty } from '@nestjs/swagger'
+import { Expose } from 'class-transformer'
+import { IsJSON, IsNotEmpty } from 'class-validator'
+
+export class FilesRoleDto {
+  @ApiProperty()
+  @Expose()
+  @IsNotEmpty()
+  @IsJSON({ each: true })
+  filesRole: string
```
imilchev, line 10: shouldn't this be an enumeration? I assume the file role cannot be some random text. Last I checked also in the migration it was stated as an enum

kachar: It's gonna be awesome if we can define the structure better, right now I see the limitation of sending the request as json string instead of proper json object, which is harder to validate properly

dplamenov: Can you give me some idea how to validate or how to define structure better?

imilchev: What about the role being an enumeration? I though Pepi was already doing something about this. Please check with him

imilchev (changes requested):

dplamenov:
@imilchev `filesRole` is something like that `[{"file": "14125.png", "role": "bg"}]`

imilchev:
> @imilchev `filesRole` is something like that `[{"file": "14125.png", "role": "bg"}]`

I understand that but that `"bg"` shouldn't be free text. It's a pre-defined set of values that can be used. Check [here](https://github.com/podkrepi-bg/api/pull/158/files#diff-8f341f62d8de4efa7d8522e95ff02fefeae465b0d553f7b8868cf46ff34a6b02).

dplamenov:
@kachar I know `CampaignFileRole` enum but how can I use for validation? More one time I say: `filesRole` is plain text(string). Thank you!

imilchev:
The type shouldn't be string but define the enumeration that is used (if not already defined) and set that as a type or something.

imilchev (approved):

PR #165, opened 2022-03-29
[expense] Created missing functionality for edit + more

### LINKED ISSUE #162: [expense] Create missing functionality for edit/update
**Expense** entity **lacks edit functionality**.
There isn't PUT or PATCH endpoint.
It needs to be added so we can edit its records at the admin panel when needed.

dpmarkov:
1. **Created endpoint** (*PATCH*) for editing an Expense record.
2. **Restricted access** to create/delete/delete all queries. It should be for authorized users only.

close #162

### DIFF
--- a/apps/api/src/expenses/expenses.controller.ts
+++ b/apps/api/src/expenses/expenses.controller.ts
@@ -1,17 +1,7 @@
-import {
-  Controller,
-  Get,
-  Post,
-  Body,
-  Param,
-  Delete,
-  Patch,
-  UnauthorizedException,
-} from '@nestjs/common'
-import { AuthenticatedUser, Public, RoleMatchingMode, Roles } from 'nest-keycloak-connect'
+import { Controller, Get, Post, Body, Param, Delete, Patch } from '@nestjs/common'
+import { Public, RoleMatchingMode, Roles } from 'nest-keycloak-connect'
 import { RealmViewSupporters, ViewSupporters } from '@podkrepi-bg/podkrepi-types'
 
-import { KeycloakTokenParsed } from '../auth/keycloak'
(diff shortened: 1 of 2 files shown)


apps/api/src/expenses/expenses.controller.ts
```diff
@@ -14,9 +25,17 @@ export class ExpensesController {
...
-  create(@Body() createExpenseDto: CreateExpenseDto) {
+  create(
+    @AuthenticatedUser()
+    user: KeycloakTokenParsed,
+    @Body()
+    createExpenseDto: CreateExpenseDto,
+  ) {
+    if (!user) {
+      throw new UnauthorizedException()
+    }
```
kachar, line 37: That's a bit too verbose if we don't actually use the user.

You might as well just skip the `@Public` decorator to require authenticated user. (the default behavior)

An important check here we should add is to make sure only admins can actually update and delete expenses.

Right now any authenticated user can update or delete any expense. That's actually quite important to be fixed @dpmarkov

kachar (changes requested):

apps/api/src/expenses/expenses.controller.ts
dpmarkov: I see, just to be sure does that snippet represent the admin roles?
```javascript
@Roles({
  mode: RoleMatchingMode.ANY,
})
```

kachar: Yes, it's exactly what we need here

dpmarkov: Sorry for the delay. It's done!

kachar (approved):

PR #166, opened 2022-03-29
Unit tests for entities

PepiPetrov:
Added unit tests for entities Beneficiary and Campaign type.

### DIFF
--- a/apps/api/src/campaign-types/campaign-types.controller.spec.ts
+++ b/apps/api/src/campaign-types/campaign-types.controller.spec.ts
@@ -35,7 +35,7 @@ const mockData = [
 describe('CampaignTypesController', () => {
   let controller: CampaignTypesController
 
-  const mockCampaignTypeservice = {
+  const mockCampaignTypeService = {
     create: jest.fn((dto) => {
       return {
         id: Date.now(),
@@ -64,7 +64,7 @@ describe('CampaignTypesController', () => {
       providers: [CampaignTypesService, PrismaService],
     })
       .overrideProvider(CampaignTypesService)
-      .useValue(mockCampaignTypeservice)
+      .useValue(mockCampaignTypeService)
       .compile()
 
     controller = module.get<CampaignTypesController>(CampaignTypesController)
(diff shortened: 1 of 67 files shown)


imilchev:
We agreed to not mock prisma with the others. Check with your colleagues on how to write tests using the real database.

imilchev:
What is the status of this PR? I see that there are still mocks

PepiPetrov:
@imilchev This is because I made them like @borislavstoychev's tests.

kachar (approved):
@imilchev I say let's move on with the mocked tests

As long as they are green for the controller and the service, I think we're good to merge it

apps/api/src/campaign-types/campaign-types.controller.spec.ts
```diff
@@ -1,21 +1,159 @@
...
+      expect(result).toHaveLength(2)
+      expect(mockCampaignTypeService.remove).toBeCalledWith('0846e9cb-0668-448b-96de-1f35dfa9a1d4')
+    })
+
+    it('should remove many items', async () => {
+      const idsToDelete = [
+        '0846e9cb-0668-448b-96de-1f35dfa9a1d4',
+        '32f16697-db01-4ae4-8269-b0d7616f9820',
+      ]
+      const result = await controller.deleteMany({ ids: idsToDelete as [string] })
```
imilchev, line 143: why is there a function to delete multiple items? that is not needed. Can we delete it? @kachar what do you thing?

imilchev, line 143: why is there a function to delete multiple items? that is not needed. Can we delete it? @kachar what do you thing?

migrations/20220329134343_added_beneficiary_photo_role/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterEnum
+ALTER TYPE "campaign_file_role" ADD VALUE 'beneficiaryPhoto';
```
imilchev, line 2: why is there a migration in a PR for tests? what happened here?

imilchev (changes requested):

apps/api/src/campaign-types/campaign-types.controller.spec.ts
kachar: We can delete it from the controller as we will not support multiple deletions at this point

apps/api/src/campaign/campaign.service.ts
```diff
@@ -143,8 +143,7 @@ export class CampaignService {
     })
 
     donations.map((donation) => {
-      if (!donation.person)
-        donation.person = { firstName: 'anonymous', lastName: '' }
+      if (!donation.person) donation.person = { firstName: 'anonymous', lastName: '' }
```
kachar, line 146:
```suggestion
      if (!donation.person) {
        donation.person = { firstName: 'anonymous', lastName: '' }
      }
```

Issue #67, opened 2022-01-16
[Campaign] Create endpoint for loading public campaign donation history


igoychev:
The API should enable loading the details of the donation made for specific campaign.

Note, that anonymous donations should be loaded with Name=anonymous. 

The dataset should be selected from PrismaDb Client using Campaign.Vaults.Donations and for now only these fields are interesting donorName, donationAmount, donationTime

kachar:
@igoychev Let's roll with `person: null` when we deal with anonymous donations. It makes it easier to check if there's a person object than to compare all places `person.firstName === 'anonymous'`

Filling in the first name and not other fields will make the person's object inconsistent.

igoychev:
Yes, I am ok with person: null. My initial description was meant to say that name 'anonymous' should be 'displayed' in the UI.

igoychev:
> @igoychev Let's roll with `person: null` when we deal with anonymous donations. It makes it easier to check if there's a person object than to compare all places `person.firstName === 'anonymous'`
> 
> Filling in the first name and not other fields will make the person's object inconsistent.

PR #178, opened 2022-03-31
Endpoint for loading campaign donation history

aironhight:
[[Campaign] Create endpoint for loading public campaign donation history #67](https://github.com/podkrepi-bg/api/issues/67)

### DIFF
--- a/apps/api/src/campaign/campaign.service.ts
+++ b/apps/api/src/campaign/campaign.service.ts
@@ -116,6 +116,39 @@ export class CampaignService {
     return this.prisma.vault.findFirst({ where: { campaignId } })
   }
 
+  async getDonationsForCampaign(campaignId: string): Promise<Donation[]> {
+    const campaign = await this.prisma.campaign.findFirst({
+      where: { id: campaignId },
+      include: {
+        vaults: true,
+      },
+    })
+
+    if (campaign === null) {
+      Logger.warn('No campaign record with id: ' + campaign)
+      throw new NotFoundException('No campaign record with id: ' + campaign)
+    }
+
+    const whereVaultIds = campaign.vaults.map((vault) => {
+      return { targetVaultId: vault.id }
+    })
+
+    const donations = await this.prisma.donation.findMany({
+      where: {
+        OR: whereVaultIds,
+      },
+      include: {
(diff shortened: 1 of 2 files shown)


kachar (approved):
This endpoint looks good @aironhight 

I was a bit curious about the use case with no available person, feels okay, but we modify db objects with hardcoded strings, maybe we need another approach to solve this.

On the frontend it's not going to be clear how to distinguish the anonymous from the real person beside checking the first name

aironhight:
I do the check for no available person, because in the DB schema it is shown as nullable.
My thoughts were - if there is no person object associated with the donation, the donation is anonymous.

image

kachar:
@aironhight Yep, that exactly right, but then why do we create a fake person and attach it to the response?

apps/api/src/campaign/campaign.service.ts
```diff
@@ -116,6 +116,40 @@ export class CampaignService {
...
+      },
+    })
+
+    donations.map((donation) => {
+      if (!donation.person)
+        donation.person = { firstName: 'anonymous', lastName: '' }
+    })
+
+    return donations
+  }
```
igoychev, line 151: the donations can be extracted from the database using a single query in this way:

```
const donations = await this.prisma.donation.findMany({
      where: { 
        status: 'succeeded',  //this is could be passed in the request too
        targetVault: {
          campaignId: { equals: campaignId}
        }
      },
      include: {
        person: { select: { firstName: true, lastName: true } },
      },
    })
```

and it generates empty array if there are no donations or no campaignId.

Example response:
`[
  {
    "id": "15cb9df5-ba2a-4d91-b409-3392a3d7348d",
    "type": "donation",
    "status": "succeeded",
    "provider": "none",
    "targetVaultId": "0ca7229e-af8b-4409-8c05-48e3318591ec",
    "extCustomerId": "",
    "extPaymentIntentId": "",
    "extPaymentMethodId": "",
    "createdAt": "2022-04-04T07:22:03.608Z",
    "updatedAt": "2022-04-04T07:30:14.787Z",
    "amount": 111,
    "currency": "BGN",
    "personId": null,
    "person": null
  }
]`

PR #182, opened 2022-04-02
181 crud operations for company entity

aironhight:


apps/api/src/company/dto/create-company.dto.ts
```diff
@@ -12,25 +13,39 @@ export class CreateCompanyDto {
 
   @Expose()
   @IsString()
-  @ApiProperty()
+  @ApiProperty({
+    description:
```
kachar, line 17: It's awesome that you've provided a nice description to the property

apps/api/src/company/dto/create-company.dto.ts
```diff
@@ -12,25 +13,39 @@ export class CreateCompanyDto {
...
+      'BULSTAT Unified Identification Code (UIC) https://psc.egov.bg/en/psc-starting-a-business-bulstat',
+  })
+  @IsUUID()
```
kachar, line 20: I'm not sure that the `companyNumber` is `UUID` format, maybe a typo

aironhight: Nope, it is not. It's unique in the DB, but not in UUID format.

kachar (approved):
Thanks for your contribution! :boat:

PR #183, opened 2022-04-03
Endpoint for admins to add donations from bank payments

aironhight:
Added endpoint enabling the admins to add donations which were transferred by bank payments.

Note: If person's First name, Last name and Email is NOT specified, the donation is going to be anonymous.

### DIFF
--- a/apps/api/src/donations/donations.controller.ts
+++ b/apps/api/src/donations/donations.controller.ts
@@ -7,6 +7,7 @@ import { DonationsService } from './donations.service'
 import { CreateSessionDto } from './dto/create-session.dto'
 import { CreatePaymentDto } from './dto/create-payment.dto'
 import { UpdatePaymentDto } from './dto/update-payment.dto'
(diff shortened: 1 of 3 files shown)


apps/api/src/donations/donations.controller.ts
```diff
@@ -67,6 +68,23 @@ export class DonationsController {
...
+    mode: RoleMatchingMode.ANY,
+  })
+  createBankPayment(
+    @AuthenticatedUser()
+    user: KeycloakTokenParsed,
+    bankPaymentDto: CreateBankPaymentDto,
+  ) {
+    if (!user) {
+      throw new UnauthorizedException()
+    }
```
kachar, line 83: Here the check for user seems redundant as the decorator `@Roles` makes sure we have an authorized user

However it's not going to fail even if we keep it that way

kachar (approved):
@aironhight Nice addition to the app

Thanks for your contribution!

PR #185, opened 2022-04-04
60 close the donation campaign once the targeted amount is reached

aironhight:
I've implemented functionality for closing the campaign if the target amount has been reached.

**NOTE**: If the target amount has been reached you could still add payments.

### DIFF
--- a/apps/api/src/campaign/campaign.service.ts
+++ b/apps/api/src/campaign/campaign.service.ts
@@ -247,6 +247,8 @@ export class CampaignService {
       })
     }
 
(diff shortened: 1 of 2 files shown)


aironhight:
I have just noticed that when you add a Donation, the vault amount does not Update.
Should we create a issue for that?

igoychev:
Well done, merging now, but please add some tests for the next PR!

aironhight:
Thanks!
I will add some tests for future PRs.

I have noticed that you can create a donation(from the donation service), which will not update the 'amount' property of the vault that it was added to.
I will create an issue for that and fix it later.

PR #189, opened 2022-04-05
186 update amount property of a vault when a donation is added

aironhight:


apps/api/src/campaign/campaign.service.ts
```diff
@@ -9,16 +9,17 @@ import {
...
 import { PrismaService } from '../prisma/prisma.service'
+import { VaultService } from '../vault/vault.service'
 import { CreateCampaignDto } from './dto/create-campaign.dto'
 import { UpdateCampaignDto } from './dto/update-campaign.dto'
 
 @Injectable()
 export class CampaignService {
-  constructor(private prisma: PrismaService) {}
+  constructor(private prisma: PrismaService,
+    @Inject(forwardRef(()=>VaultService)) private vaultService: VaultService) {}
```
kachar, line 22: This construct is quite interesting `forwardRef(()=>VaultService)` , I think we've never used it in the project

Can you elaborate on why we need it and how it helps us?

kachar (approved):
Awesome contribution @aironhight  :raised_hands: 

Thanks for digging into the donation mechanics

Can you please check the donations service and campaign controller tests prior the merge.

apps/api/src/campaign/campaign.service.ts
aironhight: We need it because we have a circular dependency (The campaign service depends on the vault service and vice versa). So, it would be impossible to compile the project without calling forwardRef. I faced some issues with this, because it is usually enough to specify a forwardRef in the module, but in this case it didn't work and was giving me the compile "circular dependency" error. So I have added the forwardRef in both the modules and the services.

kachar: Thanks for bringing this valuable knowledge to the project

aironhight:
image

Done

PR #195, opened 2022-04-08
Implement method for update user profile

dplamenov:
Implement method for update user profile

New method:
Screenshot at Apr 08 10-35-15


apps/api/src/account/account.controller.ts
```diff
@@ -1,29 +1,39 @@
...
-  getMe(
+  async getMe(
     @AuthenticatedUser()
     user: KeycloakTokenParsed,
   ) {
     if (user) {
       // Public authenticated
       console.log(user)
-      return { status: 'authenticated', user }
+      return await this.personService.findOne(user.sub as string);
```
kachar, line 23: Let's wrap the output in a variable so it's like 

```json
{
  user: {
    id: '----'
  }
}
```

instead of 

```json
{
  id:'----'
}
```


```suggestion
      return {
        user: await this.personService.findOne(user.sub as string)
      };
```

apps/api/src/account/account.service.ts
```diff
@@ -1,4 +1,12 @@
 import { Injectable } from '@nestjs/common'
+import { UpdatePersonDto } from '../person/dto/update-person.dto';
+import { PrismaService } from '../prisma/prisma.service';
 
 @Injectable()
-export class AccountService {}
+export class AccountService {
+    constructor(private prismaService: PrismaService) {}
+
+    async updateUserProfile(id: string, data: UpdatePersonDto) {
```
kachar, line 9: Let's add type-hint of the resulting object

```suggestion
    async updateUserProfile(id: string, data: UpdatePersonDto): Promise<Person> {
```

apps/api/src/person/dto/update-person.dto.ts
```diff
@@ -1,14 +1,56 @@
...
+  @Expose()
   newsletter?: boolean
+
+  @ApiProperty()
+  @Expose()
   address?: string
+  
+  @ApiProperty()
+  @Expose()
   birthday?: Date
```
kachar, line 43: Let's add decorator so we're sure that frontend sends us a correct date object

apps/api/src/account/account.controller.ts
```diff
@@ -1,29 +1,39 @@
...
     }
     // Public
     return { status: 'unauthenticated' }
   }
 
+  @Put('me')
+  async updateProfile(@AuthenticatedUser() user: KeycloakTokenParsed, @Body() data: UpdatePersonDto) {
+    if (data.birthday) {
+      data.birthday = new Date(data.birthday);
```
kachar, line 32: We may try to wrap this in a try-catch block as if a bad date is provided in `birthday` field it would result in error 500

apps/api/src/person/dto/update-person.dto.ts
```diff
@@ -1,14 +1,56 @@
...
   lastName?: string
+
+  @ApiProperty()
+  @Expose()
+  @IsString()
   email?: string
+
+  @ApiProperty()
+  @Expose()
   emailConfirmed?: boolean
```
kachar, line 23: Please remove the `emailConfirmed` field from the update person DTO.

Only the system should be able to confirm their email, people should not be allowed to confirm their email without a real email sent to them

apps/api/src/person/dto/update-person.dto.ts
```diff
@@ -1,14 +1,56 @@
...
+  @ApiProperty()
+  @Expose()
   phone?: string
+
+  @ApiProperty()
+  @Expose()
   company?: string
+  
+  @ApiProperty()
+  @Expose()
```
kachar, line 34: Let's add `@IsBoolean` decorator so we are sure that the entry data is OK

apps/api/src/person/dto/update-person.dto.ts
```diff
@@ -1,14 +1,56 @@
...
   firstName?: string
+
+  @ApiProperty()
+  @Expose()
+  @IsString()
   lastName?: string
+
+  @ApiProperty()
+  @Expose()
+  @IsString()
```
kachar, line 18: Let's switch the `@IsString()` decorator for more specific `@IsEmail` decorator.

Otherwise users will be able to add any gibberish string in the email field (if DB let's them)

kachar (changes requested):
@dplamenov Please check the comments regarding security

kachar:
Related to https://github.com/podkrepi-bg/frontend/issues/608

apps/api/src/person/dto/update-person.dto.ts
```diff
@@ -1,14 +1,55 @@
...
+  @Expose()
   personalNumber?: string
+  
+  @ApiProperty()
+  @Expose()
   keycloakId?: string
+ 
+  @ApiProperty()
+  @Expose()
   stripeCustomerId?: string
```
kachar, line 54:
```suggestion
```

kachar (approved):

kachar (approved):
Good foundation for the update in https://github.com/podkrepi-bg/frontend/pull/687

:tada:

PR #199, opened 2022-04-10
Combine two select queries in one with adding join clause in getCampaignBySlug

Bunny303:


apps/api/src/campaign/campaign.service.ts
```diff
@@ -88,15 +93,14 @@ export class CampaignService {
...
-
-    //the query always returns 1 record with the value as number or null if no donations where made yet
-    campaign['summary'] = [{ reachedAmount: reachedAmount[0]['reached_amount'] }]
+    let campaignAmountReached = 0
+    for (const vault of campaign.vaults) {
+      for (const donation of vault.donations) {
+        campaignAmountReached += donation.amount
+      }
+    }
+    campaign['summary'] = [{ reachedAmount: campaignAmountReached }]
```
kachar, line 102: Let's extract this functionality in a common function that can be reused

kachar (approved):
Thanks for the contribution @Bunny303, looks awesome!

Issue #197, opened 2022-04-10
Donation entity - Delete functionality


angelina-7:
Make delete functionality change the donation status (aka soft delete).

PR #200, opened 2022-04-10
Donation delete

angelina-7:
Delete functionality now change the donation status. 
#197

### DIFF
--- a/apps/api/src/donations/donations.service.ts
+++ b/apps/api/src/donations/donations.service.ts
@@ -145,7 +145,7 @@ export class DonationsService {
       return await this.prisma.donation.updateMany({
         where: { id: { in: ids } },
         data: {
-          status: 'deleted',
+          status: DonationStatus.deleted,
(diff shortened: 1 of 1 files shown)


apps/api/src/donations/donations.service.ts
```diff
@@ -142,8 +142,11 @@ export class DonationsService {
 
   async remove(ids: string[]): Promise<DeleteManyResponse> {
     try {
-      return await this.prisma.donation.deleteMany({
+      return await this.prisma.donation.updateMany({
         where: { id: { in: ids } },
+        data: {
+          status: 'deleted',
```
kachar, line 148: Please use the status from `DonationStatus` enum

angelina-7: Thank you, I fixed it.

kachar (approved):
@angelina-7 Thanks for the nice contribution!

PR #207, opened 2022-04-13
Restrict capaign creation

Bunny303:
Add roles for campaign create and campaign files create

### DIFF
--- a/apps/api/src/campaign-file/campaign-file.controller.ts
+++ b/apps/api/src/campaign-file/campaign-file.controller.ts
@@ -13,7 +13,8 @@ import {
 } from '@nestjs/common'
 import { FilesInterceptor } from '@nestjs/platform-express'
 import { UseInterceptors, UploadedFiles } from '@nestjs/common'
-import { Public, AuthenticatedUser } from 'nest-keycloak-connect'
+import { Public, AuthenticatedUser, RoleMatchingMode, Roles } from 'nest-keycloak-connect'
+import { RealmViewSupporters, ViewSupporters } from '@podkrepi-bg/podkrepi-types'
 import { PersonService } from '../person/person.service'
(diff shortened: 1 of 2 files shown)


apps/api/src/campaign-file/campaign-file.controller.ts
```diff
@@ -27,6 +28,10 @@ export class CampaignFileController {
 
   @Post(':campaign_id')
   @UseInterceptors(FilesInterceptor('file'))
+  @Roles({
+    roles: [RealmViewSupporters.role, ViewSupporters.role],
+    mode: RoleMatchingMode.ANY,
+  })
```
kachar, line 34: Here it seems proper to allow campaign updates only to admins so let's leave it the way you've done it :waffle:

apps/api/src/campaign/campaign.controller.ts
```diff
@@ -25,7 +25,10 @@ export class CampaignController {
   }
 
   @Post('create-campaign')
-  @Public()
+  @Roles({
+    roles: [RealmViewSupporters.role, ViewSupporters.role],
+    mode: RoleMatchingMode.ANY,
+  })
```
kachar, line 31: These actually are the admin roles and for this endpoint we want all authenticated users to be able to suggest a campaign.

To achieve that we may need only to delete the `@Public()` so the default auth guard mechanism kick in and require authenticated user.

kachar (approved):
Delightful and strict campaign creation experience

Thanks @Bunny303

apps/api/src/campaign-file/campaign-file.controller.ts
kachar: My bad, it seems we actually need this functionality in order to attach files to a campaign during the campaign creation.

However, we need to restrict the upload only to user's own campaigns to avoid security leaks.

PR #208, opened 2022-04-13
Return all donations ordered by createdAt desc

dplamenov:
Return all donations ordered by createdAt desc

https://github.com/podkrepi-bg/frontend/issues/641


kachar (approved):
Sorting go wild, thanks @dplamenov

PR #210, opened 2022-04-15
209 missing decorator body on createbankpayment and email must be without decorator isuuid in dto file

borislavstoychev:


kachar (approved):
Great catch @borislavstoychev :fishing_pole_and_fish: 

Thanks for the update!

PR #212, opened 2022-04-15
Restrict campaign creation for specific roles

Bunny303:


apps/api/src/campaign-file/campaign-file.controller.ts
```diff
@@ -13,36 +13,43 @@ import {
...
+    const person = await this.personService.findOneByKeycloakId(id)
     if (!person) {
-      Logger.warn('No person record with keycloak ID: ' + user.sub)
-      throw new NotFoundException('No person record with keycloak ID: ' + user.sub)
+      Logger.warn('No person record with keycloak ID: ' + id)
+      throw new NotFoundException('No person record with keycloak ID: ' + id)
     }
+
+    if (!isAdmin(user)) {
+      const campaign = await this.campaignService.getCampaignByIdAndPersonId(campaignId, id)
```
kachar, line 47: Adding a small note: Here we pass `id` which is `keycloakId` but the function accepts `personId` which is different ID

apps/api/src/campaign/campaign.controller.ts
```diff
@@ -25,8 +26,12 @@ export class CampaignController {
   }
 
   @Post('create-campaign')
-  async create(@Body() createDto: CreateCampaignDto) {
-    return await this.campaignService.createCampaign(createDto)
+  async create(
+    @Body() createDto: CreateCampaignDto,
+    @AuthenticatedUser() user: KeycloakTokenParsed,
+  ) {
+    const personId = !isAdmin(user) ? (user.sub as string) : undefined
```
kachar, line 33: Here the `personId` differs from `user.sub` 

The `sub` prop equals to `person.keycloakId`

kachar (approved):
Thanks for the in-depth contribution @Bunny303 

We might see if checking `resource_access` is enough or we need to check `realm_access` as well, but that's a fight for another time

Issue #213, opened 2022-04-15
[Bug] Wrong usage of sub property from type KeycloakTokenParsed


Bunny303:
When user is received from @AuthenticatedUser() user: KeycloakTokenParsed, the sub property equals to person.keycloakId.

In the cases below user.sub is used as person.id or campaign.coordinatorId and have to be fixed.
https://github.com/podkrepi-bg/api/blob/4c2e5af41ad5f80f52b9ad10a24a7c7dd10b2751/apps/api/src/donations/donations.controller.ts#L42

https://github.com/podkrepi-bg/api/blob/4c2e5af41ad5f80f52b9ad10a24a7c7dd10b2751/apps/api/src/vault/vault.controller.ts#L33
https://github.com/podkrepi-bg/api/blob/4c2e5af41ad5f80f52b9ad10a24a7c7dd10b2751/apps/api/src/vault/vault.controller.ts#L67

kachar:
Reopening the issue as only the donation controller got fixed and the vault controller is still pending

cc @dimitur2204

Bunny303:
A can fix the vault controller

PR #214, opened 2022-04-18
Get a person from the keycloakId first, send personId to request dona…

dimitur2204:
#213
## Changes
Made it so that now the `DonationsService` sends a request to the `PersonService` first to get the `personId` from the passed `keycloakId` and then sends the request for the donations.

### DIFF
--- a/apps/api/src/donations/donations.service.ts
+++ b/apps/api/src/donations/donations.service.ts
@@ -152,7 +152,7 @@ export class DonationsService {
     }
   }
 
-  async getDonationsByUser(personSub: string) {
+  async getDonationsByUser(keycloakId: string) {
     const donations = await this.prisma.donation.findMany({
       include: {
         targetVault: {
@@ -161,7 +161,7 @@ export class DonationsService {
       },
       where: {
         person: {
-          keycloakId: personSub,
+          keycloakId: keycloakId,
         },
       },
     })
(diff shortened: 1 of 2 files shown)


apps/api/src/donations/donations.service.ts
```diff
@@ -152,14 +154,15 @@ export class DonationsService {
...
+    const person = await this.person.findOneByKeycloakId(personSub)
     const donations = await this.prisma.donation.findMany({
       include: {
         targetVault: {
           include: { campaign: true },
         },
       },
-      where: { personId },
+      where: { personId: person?.id },
     })
```
Bunny303, line 166: Can you select donation by keycloakId directly without making request for person? Something like where: {person:{keycloakId: personSub}}

dimitur2204: Done in https://github.com/podkrepi-bg/api/pull/214/commits/b786da485c8a1952a097fd56db318623a7c221d5.
Good suggestion!

apps/api/src/donations/donations.service.ts
```diff
@@ -152,14 +152,18 @@ export class DonationsService {
...
     const donations = await this.prisma.donation.findMany({
       include: {
         targetVault: {
           include: { campaign: true },
         },
       },
-      where: { personId },
+      where: {
+        person: {
+          keycloakId: personSub,
```
igoychev, line 164: let's name the input parameter keycloakId instead of personSub, so that the where clause to look properly comparing same things. Also this will avoid the strangely named sub field from keycloak to propagate to our code ;)

dimitur2204: Done in https://github.com/podkrepi-bg/api/pull/214/commits/262da95dc6870e718662d88742a07fe9b821cb89

Bunny303 (approved):

PR #215, opened 2022-04-19
Donations order by person's name

angelina-7:
List of donations now will be sorted by name of donor and then by date.


kachar (approved):
In next revisions we might send the order param from the frontend so we fill more use cases.

Thanks for the contribution @angelina-7

PR #216, opened 2022-04-20
Fix check for campaign owner

Bunny303:


apps/api/src/campaign/campaign.service.ts
```diff
@@ -331,6 +351,15 @@ export class CampaignService {
     return await this.prisma.campaign.delete({ where: { id: campaignId } })
   }
 
+  async checkCampaignOwner(keycloakId: string, campaignId: string) {
+    const person = (await this.personService.findOneByKeycloakId(keycloakId)) as Person
```
kachar, line 355: @Bunny303 Why do we need to type it explicitly here with `as Person`?

apps/api/src/vault/vault.controller.ts
```diff
@@ -28,12 +28,7 @@ export class VaultController {
     @AuthenticatedUser() user: KeycloakTokenParsed,
     @Body() createVaultDto: CreateVaultDto,
   ) {
-    const campaign = await this.campaignService.getCampaignById(createVaultDto.campaignId)
-
-    if (user?.sub !== campaign.coordinatorId) {
-      throw new UnauthorizedException()
-    }
-
+    await this.campaignService.checkCampaignOwner(user.sub as string, createVaultDto.campaignId)
```
kachar, line 31: Really nice way of handling the ownership, easy to implement and validate.

apps/api/src/vault/vault.service.ts
```diff
@@ -74,6 +72,18 @@ export class VaultService {
     }
   }
 
+  async checkVaultOwner(keycloakId: string, vaultId: string) {
+    const person = (await this.personService.findOneByKeycloakId(keycloakId)) as Person
```
kachar, line 76: Same questions as above, what happens if the db cannot find that person, maybe we need to throw `UnauthorizedException` also

```suggestion
    const person = await this.personService.findOneByKeycloakId(keycloakId)
    if (!person) {
      throw new UnauthorizedException()
    }
```

kachar (commented):
@Bunny303 I've left some questions, let's discuss them

apps/api/src/campaign/campaign.service.ts
Bunny303: Person for logged user is supposed to exist in the db, because it should be created on registration. It could not exist if request for creating person fails. That will leads to scnario where user could login (registration in Keycloak is successful), but person record won't exist in our db and all of the request that are not public will fail for him.
Throwing of error when person is not found will only alert the problem. May be we could think for solution that create the person in not exists in the database. 
It's explicitly typed, because findOneByKeycloakId could return null, but in this case it's not expected to be null.

Bunny303: Ok, agree.

kachar (approved):
Thank you @Bunny303 for your swift ideas and dedication on this PR, quite an enhancement for the system.

PR #218, opened 2022-04-20
added service for custom donation list by campaign

nradkova:
`donation/recent/by-campaign/:id`
-- new endpoint and service added for query filtering donations by `ampaignId,` status - `succeeded`, sorted desc by `createdAt`.  
-- used to fix issue `#566 in frontend`

### DIFF
--- a/apps/api/src/donations/donations.controller.ts
+++ b/apps/api/src/donations/donations.controller.ts
@@ -102,4 +102,10 @@ export class DonationsController {
   ) {
     return this.donationsService.softDelete(idsToDelete)
   }
+
+  @Get('recent/by-campaign/:id')
+  @Public()
+  async getLastTenStripeDonationsByCampaign(@Param('id') id: string) {
+    return await this.donationsService.getRecentDonationsByCampaign(id)
+  }
 }
--- a/apps/api/src/donations/donations.controller.ts
+++ b/apps/api/src/donations/donations.controller.ts
@@ -103,9 +103,9 @@ export class DonationsController {
     return this.donationsService.softDelete(idsToDelete)
   }
 
-  @Get('recent/by-campaign/:id')
+  @Get('list/:campaignId')
   @Public()
-  async getLastTenStripeDonationsByCampaign(@Param('id') id: string) {
+  async getRecentDonationsByCampaign(@Param('campaignId') id: string) {
     return await this.donationsService.getRecentDonationsByCampaign(id)
   }
 }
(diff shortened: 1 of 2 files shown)


apps/api/src/donations/donations.service.ts
```diff
@@ -167,4 +167,19 @@ export class DonationsService {
     }, 0)
     return { donations, total }
   }
+
+  async getRecentDonationsByCampaign(campaignId: string) {
+    const vaults = await this.prisma.vault.findMany({ where: { campaignId } })
+    const donations = await this.prisma.donation.findMany({
+      where: {
```
Bunny303, line 174: Request to db for getting vaults could be skipped with joining vault table. Where clause could be modified 
where: { targetVault: { campaignId },  status: DonationStatus.succeeded }

apps/api/src/donations/donations.controller.ts
```diff
@@ -102,4 +102,10 @@ export class DonationsController {
   ) {
     return this.donationsService.softDelete(idsToDelete)
   }
+
+  @Get('recent/by-campaign/:id')
+  @Public()
+  async getLastTenStripeDonationsByCampaign(@Param('id') id: string) {
```
Bunny303, line 108: Name could be changed. There is no limimt set in the query for donations so it won't return only 10 if more than that exists.

apps/api/src/donations/donations.controller.ts
```diff
@@ -102,4 +102,10 @@ export class DonationsController {
   ) {
     return this.donationsService.softDelete(idsToDelete)
   }
+
+  @Get('recent/by-campaign/:id')
```
Bunny303, line 106: Route could be 'list/:campaignId' or 'by-campaign/:campaignId'.

apps/api/src/donations/donations.service.ts
nradkova: Thank you for the optimization.

apps/api/src/donations/donations.controller.ts
nradkova: This is really a typo as it was initially limited to 10 results. Thanks.

Bunny303: Actually will be good to have limit set in the query based on the needs of the front end.

nradkova: In the requirements is set to be shown all benefactors and donations amount -- `Виж всички`.  That's why the query is changed to take all of them. I left the limitation in the front end with a comment, in order to be discussed.

PR #219, opened 2022-04-21
Local Keycloak

borislavstoychev:
Please change ``` .env.local``` on frontend to


docker-compose.yml
```diff
@@ -60,6 +60,29 @@ services:
...
+  keycloak:
+    container_name: keycloak
+    image: docker.io/bitnami/keycloak:15
+    environment:
+      - KEYCLOAK_USER=admin
+      - KEYCLOAK_PASSWORD=admin
+      - DEBUG=true
+      - DEBUG_PORT='*:8787'
+      - DB_VENDOR=POSTGRES
+      - DB_ADDR=pg-db
```
kachar, line 72: I'm not sure it's a good idea to reuse the same db for backend and keycloak

Let's make another db instance dedicated to keycloak

docker-compose.yml
```diff
@@ -60,6 +60,29 @@ services:
...
+      - TZ=Europe/Sofia
+    depends_on:
+      - pg-db
+    ports:
+      - '9080:8080'
+      - '8787:8787' # debug port
+    networks:
+      - backend-net
+
   # Flyway is deprecated, Prisma migrations are taking place
```
kachar, line 86: We can remove all Flyway commended code in the same PR

docker-compose.yml
```diff
@@ -60,6 +60,29 @@ services:
     networks:
       - backend-net
 
+  keycloak:
+    container_name: keycloak
+    image: docker.io/bitnami/keycloak:15
```
kachar, line 65:
```suggestion
    # https://github.com/bitnami/bitnami-docker-keycloak
    image: docker.io/bitnami/keycloak:15
```

kachar:
Let's consider adding our custom theme as described at https://github.com/bitnami/bitnami-docker-keycloak#adding-custom-themes

borislavstoychev:
There is still the same 401 error unauthorized. I wonder if these warnings are the problem.
image

Bunny303:
Did you add KEYCLOAK_SECRET in the api?

borislavstoychev:
@Bunny303  you mean this secret key
image

borislavstoychev:
image

borislavstoychev:
@kachar  @Bunny303  the error was in my ```.env.local```  file on api. I had set KEYCLOAK_SECRET there and I probably never used this one in ```.env``` file.

igoychev:
> Can we remove required email verification from dev realm? It will skip a step in user registration locally.

Yep, we can switch the verification off, and I was also thinking to import some default users for the main roles with the realm file, so that all comes ready-to-use. Will check how that can happen...

kachar:
Great work!

Can we import multiple files?
If we can we might have one file that includes the real options and one file that includes the dev overrides and test users

quantum-grit:
> Great work!
> 
> Can we import multiple files? If we can we might have one file that includes the real options and one file that includes the dev overrides and test users

Yes, multiple files per realm are possible, though I couldn't make it using the EXTRA-ARGS of the main keycloak container. It can be done with another container called keycloak-config-cli which allows realm import and updates from multiple files - see the commented section in docker-compose.yaml. 

Anyway, I did it with a single file without this new container to not load the dev machines, and we can split and extend easily when more complex config comes in place.

kachar (approved):
Marvelous work @borislavstoychev  and @igoychev :rocket: :rocket: :rocket:

kachar:
> Can we remove required email verification from dev realm? It will skip a step in user registration locally.

It's done! Thanks for the suggestion @Bunny303

PR #222, opened 2022-04-26
Add calculation of unique donors in getting campaign

Bunny303:


kachar (approved):
Thanks, @Bunny303 for solving this one out

This calculation seems to grow and we might thing of pure SQL solution if this trend continues.

PR #224, opened 2022-04-28
Add new route for create registered user

Bunny303:


kachar (approved):
Marvelous upgrade @Bunny303 

Thanks for fixing the `person.id != keycloakId` one more time

PR #231, opened 2022-04-30
Restricted public access

Bunny303:


apps/api/src/campaign/campaign.controller.ts
```diff
@@ -28,7 +28,6 @@ export class CampaignController {
   ) {}
 
   @Get('list')
-  @Public()
```
kachar, line 31: We need to leave this one public as it used on the main campaigns page.

apps/api/src/expenses/expenses.controller.ts
```diff
@@ -11,12 +11,19 @@ export class ExpensesController {
   constructor(private readonly expensesService: ExpensesService) {}
 
   @Public()
```
kachar, line 13: It should be either `@Public()` or `@Roles()`

I'm not certain about the behavior of both together.

```suggestion
```

apps/api/src/recurring-donation/recurring-donation.controller.ts
```diff
@@ -11,19 +11,20 @@ export class RecurringDonationController {
...
+  @Roles({
+    roles: [RealmViewSupporters.role, ViewSupporters.role],
+    mode: RoleMatchingMode.ANY,
+  })
   findAll() {
     return this.recurringDonationService.findAll()
   }
 
   @Get(':id')
-  @Public()
```
kachar, line 26: I don't find reasons to have this endpoint available to authenticated users. Maybe we can restrict it to the admins as well

apps/api/src/campaign/campaign.controller.ts
Bunny303: In the main campaign page should be returned only active and completed campaigns. I thought this route to be used for admin list with campaigns and new one to be created for the main page.

Bunny303: Actually in the existing one should be added a filter for active and completed and new route with admin access only could be added.

kachar (approved):
Outstanding work @Bunny303  Now the api feels much more secure :shield: :closed_lock_with_key:

PR #233, opened 2022-05-02
Added new route for fetching all campaigns

Bunny303:


apps/api/src/campaign/campaign.controller.ts
```diff
@@ -33,6 +33,12 @@ export class CampaignController {
     return this.campaignService.listCampaigns()
   }
 
+  @Get('list-all')
+  @Public()
+  async getAdminList() {
+    return this.campaignService.listAllCampaigns()
+  }
+
```
igoychev, line 41: only AdminRole should be able to see all campaigns

PR #234, opened 2022-05-05
230 restric create donations only for active or completed campaigns

Bunny303:
Added property in campaign model that allows donations to be done after terget amount is reached


kachar (approved):
A clever approach to the problem, thanks @Bunny303

PR #237, opened 2022-05-07
Added lib and function for generating hash

Bunny303:


kachar (approved):
Nice implementation @Bunny303 I hope the custom dictionary will cover our needs.

PR #239, opened 2022-05-11
Campaign-report and report-file api

nradkova:
- ReportFile model created
- CampaignReport model - updated - added notifierType and relation to ReportFile
- controllers and services for both entities are implemented

### DIFF
--- a/apps/api/src/app/app.module.ts
+++ b/apps/api/src/app/app.module.ts
@@ -43,6 +43,7 @@ import { WithdrawalModule } from '../withdrawal/withdrawal.module'
 import { CampaignTypesModule } from '../campaign-types/campaign-types.module'
 import { RecurringDonationModule } from '../recurring-donation/recurring-donation.module'
 import { TransferModule } from '../transfer/transfer.module'
+import { ReportFileModule } from '../report-file/report-file.module'
 
 @Module({
   imports: [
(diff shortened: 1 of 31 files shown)


kachar:
The report file brings so much complexity

Is it really a mandatory part of the flow?

nradkova:
@kachar It is predicted that a person can report and can attach files as a prove. Maybe it will be better to remove campaign-report from the support module so that the dependencies and complexity will be lower.

nradkova:
@kachar , please check the new module for the campaign report. I have moved it from the support module to a separate one.

Bunny303:
Is it an option campaign file to be reused for report file?

kachar:
Looks like we duplicate a lot of logic for uploading/storing a file

We might extract the common parts and reuse them for campaign-file and report-file

@Bunny303 is that what you're referring to?

Bunny303:
@kachar Yes, to extract common logic and reuse it. For types also could be extracted common properties.

nradkova:
@kachar, I do not know how to fix this GitGuardian Security Check. I reset all commits from today. Yesterday all checks passed successfully (today I just deleted some commented code in the report-file folder).  The problem is set to be in `Merge remote-tracking branch 'origin/master' into campaign-report-api -[c24d5fb]` - 11 days ago. Currently, my branch is up to date with `master`. As far as I can see the problem is in the `readme` and `env.local.example`, `manifests/keycloak-master-realm.json` files with Keycloak-Secret shown, but they are the same as in master (the secret is written there as well). Please, give me some guidance on how to fix the problem.

kachar:
@nradkova Thanks for bringing that up, indeed the git guardian check is triggered by the code in master branch. There's nothing we can do right now and there's no expectation that you fix the related code. You can freely focus on the feature instead.

Issue #168, opened 2022-03-29
Authentication token doesn't refresh after expiration


Kalin-Konstantinov:
There is a problem with the refresh token and it does not refresh after expiration. Also, when the token is not valid, it is not removed.

PR #244, opened 2022-05-16
Refresh token endpoint

dimitur2204:
Related to #168, but I can't quite understand the issue!

Added an endpoint for fetching an `access_token` when you give a `refresh_token`.
image

I needed to use a normal http request since the Keycloak SDK does not support that request or at least I couldn't find it.
Because of that I needed to add `@nestjs/axios` to the dependencies

### DIFF
--- a/apps/api/src/auth/auth.service.ts
+++ b/apps/api/src/auth/auth.service.ts
@@ -66,10 +66,10 @@ export class AuthService {
     const clientId = this.config.get<string>('keycloak.clientId')
     const tokenUrl = `${this.config.get<string>('keycloak.serverUrl')}/realms/${this.config.get<string>('keycloak.realm')}/protocol/openid-connect/token`
     const data = {
-      'client_id':clientId as string,
-      'client_secret':secret as string,
-      'refresh_token': refreshDto.refreshToken,
-      'grant_type':'refresh_token'
+      client_id: clientId as string,
+      client_secret: secret as string,
+      refresh_token: refreshDto.refreshToken,
+      grant_type: <const>'refresh_token'
     }
     const params = new URLSearchParams(data)
     return this.httpService.post<KeycloakErrorResponse | TokenResponseRaw>(tokenUrl,params.toString()).pipe(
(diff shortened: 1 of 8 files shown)


apps/api/src/auth/dto/refresh.dto.ts
```diff
@@ -0,0 +1,11 @@
+import { ApiProperty } from '@nestjs/swagger'
+import { Expose } from 'class-transformer'
+import { IsNotEmpty, IsString } from 'class-validator'
+
+export class RefreshDto {
+  @ApiProperty()
+  @Expose()
+  @IsNotEmpty()
+  @IsString()
+  public readonly refreshToken: string
```
kachar, line 10: We can add here `@IsJWT` to be sure that the format is good

apps/api/src/auth/auth.service.ts
```diff
@@ -52,6 +61,28 @@ export class AuthService {
...
+  async issueTokenFromRefresh(refreshDto: RefreshDto) {
+    const secret = this.config.get<string>('keycloak.secret')
+    const clientId = this.config.get<string>('keycloak.clientId')
+    const tokenUrl = `${this.config.get<string>('keycloak.serverUrl')}/realms/${this.config.get<string>('keycloak.realm')}/protocol/openid-connect/token`
+    const data = {
+      'client_id':clientId as string,
+      'client_secret':secret as string,
+      'refresh_token': refreshDto.refreshToken,
+      'grant_type':'refresh_token'
+    }
```
kachar, line 73:
```suggestion
    const data = {
      client_id: clientId as string,
      client_secret: secret as string,
      refresh_token: refreshDto.refreshToken,
      grant_type: <const>'refresh_token'
    }
```

igoychev:
Guys I see this was merged and I'm too late to the game:
Yet, could you check if the refresh token endpoint mentioned here could help:
https://stackoverflow.com/questions/51386337/refresh-access-token-via-refresh-token-in-keycloak
also for future, it looks all endpoints are listed here:
https://keycloak.podkrepi.bg/auth/realms/webapp/.well-known/openid-configuration

kachar:
@igoychev Thanks for the input, indeed we're doing exactly the same as you mentioned - sending a request to `/protocol/openid-connect/token`

Both wrapper libraries `keycloak-connect` and `@keycloak/keycloak-admin-client`` does not provide wrapper method for this call so we're using the native rest call.

https://github.com/podkrepi-bg/api/blob/3850bcba2b88f3e13078f39e0cf8a2c37f590d46/apps/api/src/auth/auth.service.ts#L65-L83

PR #248, opened 2022-05-20
[Feature] Login from external Identity Providers

dimitur2204:
Add a call to be able to get a `keycloakAccess` token through a third party provider.
Add a `picture` to the `Person` entity that can be taken from the identity providers.

### DIFF
--- a/apps/api/src/auth/auth.service.ts
+++ b/apps/api/src/auth/auth.service.ts
@@ -99,6 +99,7 @@ export class AuthService {
         email: userInfo.email || '',
         firstName: userInfo.given_name || '',
         lastName: userInfo.family_name || '',
+        picture: providerDto.picture,
       },
       // Store keycloakId to the person with same email
       update: { keycloakId: userInfo.sub },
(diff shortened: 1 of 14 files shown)


dimitur2204:
Tests are not working currently.

dimitur2204:
> Tests are not working currently.

Fixed

apps/api/src/auth/auth.service.ts
```diff
@@ -92,6 +87,51 @@ export class AuthService {
...
+        firstName: userInfo.given_name || '',
+        lastName: userInfo.family_name || '',
+        picture: providerDto.picture,
+        keycloakId: userInfo.sub,
+      },
+      // Store keycloakId to the person with same email
+      update: { keycloakId: userInfo.sub },
+      where: { email: userInfo.email },
+    })
+    return $tokenObs
```
dimitur2204, line 117: There might be a better way to do all these with `rxjs` operators, but I am not the best there. If you have any suggestions on how to do that I am open

kachar (approved):
Thanks for the heavy-lifting session @dimitur2204 
Works good, let's check it out on staging

PR #254, opened 2022-06-06
Campaign creation improvements

quantum-grit:
## Motivation and context
Trying to polish the campaign creation and fixing small things all over 

- corrected the comment in migration - it is safe to execute on existing data
- list campaigns now returns soonest campaign by endDate on top
- added logging for campaign creation
- replaced short-unique-id with nanoid as it was already in golevelup/nestjs-stripe
- renamed bankHash to paymentReference to better illustrate what it is needed for
- extended the paymentReference code from 5 to 12 symbols and added dashes to make it friendlier. Codes look like this now: "NY5P-KVO4-DNBZ" and guarantee 35years until 1% probability of duplicate
- paymentReference is now unique per campaign in database schema 
- all authenticated users should be able to create campaigns in draft mode. 
- added: campaign.state in campaign request so that it can be changed in frontend

### DIFF
--- a/apps/api/src/campaign/dto/create-campaign.dto.ts
+++ b/apps/api/src/campaign/dto/create-campaign.dto.ts
@@ -12,7 +12,7 @@ import {
 } from 'class-validator'
 import { ApiProperty } from '@nestjs/swagger'
 import { Expose, Type } from 'class-transformer'
-import { Currency, Prisma } from '.prisma/client'
+import { CampaignState, Currency, Prisma } from '.prisma/client'
 import { getPaymentReference } from '../helpers/payment-reference'
 
 @Expose()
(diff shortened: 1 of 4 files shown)


apps/api/src/campaign/dto/create-campaign.dto.ts
```diff
@@ -98,6 +98,13 @@ export class CreateCampaignDto {
   @Type(() => Date)
   endDate: Date | null
 
+  @ApiProperty()
+  @Expose()
+  @IsOptional()
+  @IsString()
+  @IsEnum(CampaignState)
+  state: CampaignState | undefined
```
kachar, line 106: When we add the control of the state of a campaign to the frontend we open the gates for intervention and for example submit a campaign as `approved`.

@quantum-grit Have you considered this when you've updated the code?

PR #255, opened 2022-06-06
keycloak initialization fixes

quantum-grit:
## Motivation and context
After testing complete reset of local docker it turned out Keycloak v18 does not initialize and does not import realm at all.

Changes:
- changed keycloak version to v17 because v18 does not initialize db and does not import realm on clean setup
- added uuids for default users when importing realm
- updated person.seed to create default users with same keycloakIds as in keycloak-webapp-realm file


apps/api/src/campaign/dto/create-campaign.dto.ts
```diff
@@ -32,11 +32,6 @@ export class CreateCampaignDto {
   @IsString()
   essence: string
 
-  @ApiProperty()
-  @Expose()
-  @IsString()
-  bankHash: string
```
kachar, line 38: Good catch, we don't need the payment reference in the input for campaign creation

schema.prisma
```diff
@@ -156,7 +156,7 @@ model Campaign {
   slug                    String           @unique @db.VarChar(250)
   title                   String           @db.VarChar(200)
   essence                 String           @db.VarChar(500)
-  bankHash                String           @map("bank_hash") @db.VarChar(5)
+  paymentReference        String           @map("payment_reference") @unique @db.VarChar(15)
```
kachar, line 159: You've picked a more suitable and appropriate name for the field, thanks!

Uniqueness of the field is a must-have :+1:

kachar (approved):
@quantum-grit Awesome increment!

Thanks for adding guiding comments

migrations/20220601201014_rename_campaign_report_to_irregularity/migration.sql
```diff
@@ -1,9 +1,5 @@
 /*
-  Warnings:
-
-  - You are about to drop the `campaign_reports` table. If the table is not empty, all the data it contains will be lost.
-  - You are about to drop the `report_files` table. If the table is not empty, all the data it contains will be lost.
-
+The below migration is written by hand because the default migration drops and recreates tables
```
kachar, line 2: @quantum-grit Modifying already deployed migration changes it's file sha hash, so a warning will be shown in the logs when we deploy further migrations.

This is general info, no actions are needed.

Bunny303:
Sorry for late comments :) I've added them yesterday, but didn't cklick on submit review.

apps/api/src/campaign/campaign.controller.ts
```diff
@@ -54,11 +54,18 @@ export class CampaignController {
...
+    }
+
+    let person
+    if (!createDto.coordinatorId) {
+      //Find the creator personId and make him a coordinator
       person = await this.personService.findOneByKeycloakId(user.sub as string)
+
       if (!person) {
-        throw new NotFoundException('No person found for logged user')
+        Logger.error('No person found in database for logged user: ' + user.name)
```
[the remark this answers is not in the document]
kachar: The field `email` is also unique both in our db and keycloak db, but it might be considered sensitive data.
I do agree we can improve the log, but that should not be a blocker

PR #256, opened 2022-06-06
Edit user profile

nradkova:
- New endpoints and services in the account module in order to enable users to change their password and personal info, as well as to delete/disable their account. 
- The functionality of updating users is for both PersonService and AuthService.
- The delete account request does not actually delete the user in Keycloak, just makes him `enabled:false`.


apps/api/src/account/account.controller.ts
```diff
@@ -34,7 +35,45 @@ export class AccountController {
     @AuthenticatedUser() user: KeycloakTokenParsed,
     @Body() data: UpdatePersonDto,
   ) {
-    return await this.accountService.updateUserProfile(user.sub as string, data)
+    let person: Person | null = null
+    try {
+      person = await this.accountService.updateUserProfile(user, data)
+    } catch (err) {
+      Logger.warn(`Failed to update user with keycloakId ${user.sub}`)
```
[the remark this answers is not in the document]
nradkova: Thanks for your comment. I've fixed the errors.

apps/api/src/account/account.service.ts
```diff
@@ -1,13 +1,30 @@
...
 @Injectable()
 export class AccountService {
-  constructor(private prismaService: PrismaService) {}
+  constructor(private authService: AuthService) {}
+
+  async updateUserProfile(
+    user: KeycloakTokenParsed,
+    data: UpdatePersonDto,
+  ): Promise<Person | null> {
+    return await this.authService.updateUser(user.sub as string, data)
```
igoychev, line 17: let's make user.sub mandatory field in KeycloakTokenParsed type, so that we don't need to cast it like this

apps/api/src/auth/auth.service.ts
```diff
@@ -220,4 +221,52 @@ export class AuthService {
       where: { email: registerDto.email },
     })
   }
+
+  async updateUser(id: string, updateDto: UpdatePersonDto) {
```
igoychev, line 225: please rename id to keycloakId so that it instructs which id is expected here

apps/api/src/auth/auth.service.ts
```diff
@@ -220,4 +221,52 @@ export class AuthService {
...
+      data: {
+        firstName: updateDto.firstName,
+        lastName: updateDto.lastName,
+        email: updateDto.email,
+        birthday: updateDto.birthday,
+      },
+    })
+  }
+
+  async updateUserPassword(id: string, credentialsDto) {
```
igoychev, line 250: please rename id to keycloakId so that we know which id is expected here

apps/api/src/auth/auth.service.ts
```diff
@@ -220,4 +221,52 @@ export class AuthService {
...
+      credential: {
+        temporary: false,
+        type: 'password',
+        value: credentialsDto.password,
+      },
+    })
+    return true
+  }
+
+  async disableUser(id: string) {
```
igoychev, line 263: please rename id to keycloakId so that it to show which id is expected here

igoychev (changes requested):
well done!

apps/api/src/account/account.service.ts
nradkova: It is used in some other controllers, shall I change it there as well?

igoychev: yes, let's be consistent

apps/api/src/account/account.controller.ts
```diff
@@ -34,7 +35,45 @@ export class AccountController {
...
+      person = await this.accountService.updateUserProfile(user, data)
+    } catch (err) {
+      Logger.error(`Failed to update user with keycloakId ${user.sub}. Error is: ${err}`)
+      throw err
+    }
+
+    return {
+      user: person,
+    }
+  }
```
kachar, line 49: The following will have the same effect but with less type definitions:

```suggestion
    try {
      const user = await this.accountService.updateUserProfile(user, data)
      return { user }
    } catch (err) {
      Logger.error(`Failed to update user with keycloakId ${user.sub}. Error is: ${err}`)
      throw err
    }
  }
```

apps/api/src/auth/auth.service.ts
```diff
@@ -220,4 +221,52 @@ export class AuthService {
...
+    await this.admin.users.update(
+      { id: keycloakId },
+      {
+        username: updateDto.email,
+        email: updateDto.email,
+        firstName: updateDto.firstName,
+        lastName: updateDto.lastName,
+        emailVerified: true,
+        requiredActions: [],
+        enabled: true,
```
kachar, line 236: I'm not sure if `updateUser` should set the `enable` flag to true. 

If we want to mark an account as disabled they will have the option to update their name and get enabled again by themselves.

kachar (approved):
Very detailed work @nradkova! Thanks for your contribution!

igoychev (approved):

PR #259, opened 2022-06-12
fixing admin create beneficiery issue 818

quantum-grit:
Fixing issues related to Admin Crete Beneficiary Form https://github.com/podkrepi-bg/frontend/issues/818
Additional fixes added as noticed while debuging the campaing creation

- fixed: in create/update Beneficiary the personId can be optional, if passing companyId and vice versa
- rename: getCampaignByIdAndPersonId to getCampaignByIdAndCoordinatorId to better reflect that this method is checking if the authenticated user is coordinator of the campaign
- fixed: keycloak.isAdmin function to check for realmRoles too
- fixed: expenses list method was not exposed via API

### DIFF
--- a/apps/api/src/auth/keycloak.ts
+++ b/apps/api/src/auth/keycloak.ts
@@ -1,11 +1,14 @@
+import { Logger } from '@nestjs/common'
 import { RealmViewSupporters, ViewSupporters } from '@podkrepi-bg/podkrepi-types'
 
 export function isAdmin(user: KeycloakTokenParsed): boolean {
-  if (!user.resource_access) {
+  Logger.debug('User info: ' + JSON.stringify(user))
+
+  if (!(user.resource_access || user.realm_access)) {
     return false
   }
 
-  const roles = [RealmViewSupporters.role, ViewSupporters.role].map((role) => {
+  const adminRoles = [RealmViewSupporters.role, ViewSupporters.role].map((role) => {
     const [key, roleName] = role.split(':')
     return {
       key,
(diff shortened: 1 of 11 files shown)


apps/api/src/auth/keycloak.ts
```diff
@@ -1,21 +1,36 @@
+import { Logger } from '@nestjs/common'
 import { RealmViewSupporters, ViewSupporters } from '@podkrepi-bg/podkrepi-types'
 
 export function isAdmin(user: KeycloakTokenParsed): boolean {
```
kachar, line 4: This function becomes quite complex, maybe we need to add a unit test to confirm all the expected cases

apps/api/src/beneficiary/dto/create-beneficiary.dto.ts
```diff
@@ -14,8 +22,26 @@ export class CreateBeneficiaryDto {
   @IsNotEmpty()
   @Expose()
   @IsUUID()
+  @IsOptional()
+  @ValidateIf(
+    (data) =>
+      (data.type === BeneficiaryType.individual && data.personId) ||
+      (data.type === BeneficiaryType.company && data.companyId),
+  )
```
kachar, line 30: Nice addition! Great usage of the `@ValidateIf` decorator

apps/api/src/campaign/campaign.service.ts
```diff
@@ -78,9 +78,12 @@ export class CampaignService {
...
   }
 
-  async getCampaignByIdAndPersonId(campaignId: string, personId: string): Promise<Campaign | null> {
+  async getCampaignByIdAndCoordinatorId(
+    campaignId: string,
+    coordinatorId: string,
+  ): Promise<Campaign | null> {
     const campaign = await this.prisma.campaign.findFirst({
-      where: { id: campaignId, coordinator: { personId } },
+      where: { id: campaignId, coordinator: { personId: coordinatorId } },
```
kachar, line 86: The `coordinatorId` is different from `personId` - two separate models and two separate tables.

We need to keep in mind that `Coordinatr` model has own field `personId` inside of it.

kachar (approved):
Feels promising, right on point

PR #265, opened 2022-06-14
817 featureadmin add donated amount column and link to donations to campaign grid

borislavstoychev:
[frontend](https://github.com/podkrepi-bg/frontend/pull/822)

### DIFF
--- a/apps/api/src/campaign/campaign.service.ts
+++ b/apps/api/src/campaign/campaign.service.ts
@@ -57,7 +57,7 @@ export class CampaignService {
   }
 
   async listAllCampaigns(): Promise<Campaign[]> {
-    return await this.prisma.campaign.findMany({
+    const campaigns = await this.prisma.campaign.findMany({
       orderBy: {
         endDate: 'asc',
       },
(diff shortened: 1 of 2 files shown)


apps/api/src/campaign/campaign.service.ts
```diff
@@ -57,17 +57,29 @@ export class CampaignService {
...
         campaignType: { select: { name: true } },
         beneficiary: { select: { person: { select: { firstName: true, lastName: true } } } },
         coordinator: { select: { person: { select: { firstName: true, lastName: true } } } },
-        vaults: { select: { id: true, amount: true } },
+        vaults: { select: { donations: { select: { amount: true } }, amount: true } },
+        incomingTransfers: { select: { amount: true } },
+        outgoingTransfers: { select: { amount: true } },
       },
     })
+    for (const campaign of campaigns) {
```
kachar, line 73: @borislavstoychev Have you checked if you can reuse `addReachedAmountAndDonors` method?

apps/api/src/campaign/campaign.service.ts
```diff
@@ -57,17 +57,29 @@ export class CampaignService {
...
+      let campaignAmountReached = 0
+      for (const vault of campaign.vaults) {
+        for (const donation of vault.donations) {
+          campaignAmountReached += donation.amount
+        }
+      }
+      campaign['reachedAmount'] = campaignAmountReached
```
kachar, line 80: The frontend expects that `reachedAmount` is part of `campaign.summary.reachedAmount`

Do we need to update it there as well?

apps/api/src/campaign/campaign.service.ts
igoychev: I agree with Ilko here - please use the calculated totals on server side, and no need to return all separate amounts to frontend

PR #266, opened 2022-06-15
Added bank transaction bucket as env

kachar:
Added bank transaction bucket as env `S3_BUCKET_BANK_TRANSACTION`

The default value is `banktransactions-files` for dev/prod usage as we upload at different S3 endpoints (`https://cdn-dev.podkrepi.bg` vs `https://cdn.podkrepi.bg`)

### DIFF
--- a/.env
+++ b/.env
@@ -33,6 +33,7 @@ DATABASE_URL=postgres://postgres:postgrespass@localhost:5432/postgres?schema=api
 S3_ENDPOINT=https://cdn-dev.podkrepi.bg
 S3_ACCESS_KEY=
 S3_SECRET_ACCESS_KEY=
+S3_BUCKET_BANK_TRANSACTION=banktransactions-files
 
 ## Keycloak ##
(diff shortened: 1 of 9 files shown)


.env
```diff
@@ -33,6 +33,7 @@ DATABASE_URL=postgres://postgres:postgrespass@localhost:5432/postgres?schema=api
 S3_ENDPOINT=https://cdn-dev.podkrepi.bg
 S3_ACCESS_KEY=
 S3_SECRET_ACCESS_KEY=
+S3_BUCKET_BANK_TRANSACTION=banktransactions-files
```
igoychev, line 36: IMHO the bucket names shouldn't be specified with the environment variables, because they are the same concept like the tables of a database, which we don't put in .env file.

On the other hand, the S3 access is ok to be in the .env in the same way the Database server access is.

kachar: Alright, closing the PR then

PR #269, opened 2022-06-16
Feature/filter donations

stann1:
The default sort order of donations was also changed a bit, because it did not make sense from business perspective.

### DIFF
--- a/apps/api/src/donations/donations.service.ts
+++ b/apps/api/src/donations/donations.service.ts
@@ -76,13 +76,14 @@ export class DonationsService {
     ]
   }
 
+  /**
+   * Lists all donations
+   * @param campaignId (Optional) Filter by campaign id
+   */
   async listDonations(campaignId?: string): Promise<Donation[]> {
     if (campaignId) {
       return await this.prisma.donation.findMany({
         where: { targetVault: {campaign: {id: campaignId}} },
-        include: {
-          targetVault: true,
-        },
         orderBy: [{ createdAt: 'desc' }],
       })
     } else {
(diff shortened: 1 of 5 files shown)


apps/api/src/donations/donations.service.ts
```diff
@@ -76,10 +76,21 @@ export class DonationsService {
...
+    if (campaignId) {
+      return await this.prisma.donation.findMany({
+        where: { targetVault: {campaign: {id: campaignId}} },
+        orderBy: [{ createdAt: 'desc' }],
+      })
+    } else {
+      return await this.prisma.donation.findMany({
+        orderBy: [{ createdAt: 'desc' }],
+      })
+    }
```
igoychev, line 93: Just tested: Here there is no need of the if condition, the where clause will return all if the id is undefined.
so leave only this code:
```
return await this.prisma.donation.findMany({
        where: { targetVault: {campaign: {id: campaignId}} },
        orderBy: [{ createdAt: 'desc' }],
      })
```

docker-compose.yml
```diff
@@ -49,7 +49,7 @@ services:
       - pg-db-data:/bitnami/postgresql
 
     ports:
-      - 5432:5432
+      - '${DB_PORT?}:5432'
```
igoychev, line 52: good one too!

igoychev (changes requested):
great improvements!
thanks for adding the .vscode config and showing the APIQuery style. Professional function description too!

apps/api/src/donations/donations.service.ts
stann1: not exactly - if the value is an empty string then there will be an exception. While the routing middleware correctly converts empty string to undefined, this business logic can be reused from anywhere in the code, where such conversion will not be available

kachar: I think we're good to continue as is with the `campaignId` param

PR #271, opened 2022-06-20
created the banktransactions-files bucket

quantum-grit:
Created the banktransactions-files bucket and renamed test bucket to it.
Also updated README.md with instructions for future bucket creation.


kachar (approved):
Thanks for the update @quantum-grit

kachar:
@quantum-grit Have you created the bucket on staging and production or it should be part of the deployment procedure?

PR #280, opened 2022-07-06
Enhancement of vault transactions and operations

stann1:
this is working but still requires a lot of polishing, which will be done as follow-up issues:
- separate update and approval functions in the services
- homogenize the status enums for transfers/withdrawals/expenses, as well as the endpoint names and verb types in the controllers
- remove unnecessary props from update dtos, because when approving - only reason and status are sufficient
- handle concurrency exceptions


apps/api/src/transfer/transfer.service.ts
```diff
@@ -10,12 +10,28 @@ import { UpdateTransferDto } from './dto/update-transfer.dto'
...
-    } catch (error) {
-      throw new NotFoundException(error)
+    const sourceVault = await this.prisma.vault.findFirst({
+      where: {
+        id: createTransferDto.sourceVaultId,
+      },
+      rejectOnNotFound: true,
+    })
+
+    if (sourceVault.amount - sourceVault.blockedAmount - createTransferDto.amount <= 0) {
```
igoychev, line 24: same as before: it should be possible to go down to 0

apps/api/src/withdrawal/withdrawal.service.ts
```diff
@@ -8,8 +8,27 @@ import { UpdateWithdrawalDto } from './dto/update-withdrawal.dto'
...
+    const vault = await this.prisma.vault.findFirst({
+      where: {
+        id: createWithdrawalDto.sourceVaultId,
+      },
+      rejectOnNotFound: true,
+    })
+    if (vault.amount - vault.blockedAmount - createWithdrawalDto.amount <= 0) {
+      throw new BadRequestException('Insufficient amount in vault.')
+    }
+
```
igoychev, line 24: it should be possible to go down to 0

apps/api/src/withdrawal/withdrawal.service.ts
```diff
@@ -30,16 +49,70 @@ export class WithdrawalService {
...
+    } else if (
+      updateWithdrawalDto.status === WithdrawStatus.declined ||
+      updateWithdrawalDto.status === WithdrawStatus.cancelled
+    ) {
+      // in case of rejection: unblock amount
+      writeVault = this.prisma.vault.update({
+        where: { id: vault.id },
+        data: { blockedAmount: {decrement: withdrawal.amount} },
+      })
+    }
```
igoychev, line 101: to avoid the if, can initialize above with the blockedAmount decrement

igoychev (changes requested):
Well done and great example for first usage of transactions on the project!
Please check for the condition to going down to 0 amount.

apps/api/src/transfer/transfer.service.ts
```diff
@@ -45,16 +61,87 @@ export class TransferService {
...
+    } else if (
+      updateTransferDto.status === TransferStatus.declined ||
+      updateTransferDto.status === TransferStatus.cancelled
+    ) {
+      // in case of rejection: unblock amount
+      writeSrcVault = this.prisma.vault.update({
+        where: { id: srcVault.id },
+        data: { blockedAmount: {decrement: transfer.amount} },
+      })
+    }
```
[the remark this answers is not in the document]
stann1: same comment as before...

apps/api/src/withdrawal/withdrawal.service.ts
stann1: same comment as before...

igoychev (approved):

PR #290, opened 2022-07-13
Included blockedAmound into the campaign List and remove unused fields

borislavstoychev:
Link to issue in [frontend](https://github.com/podkrepi-bg/frontend/issues/850)


apps/api/src/campaign/campaign.service.ts
```diff
@@ -71,15 +71,12 @@ export class CampaignService {
...
+            blockedAmount: true,
           },
         },
-        incomingTransfers: { select: { amount: true } },
-        outgoingTransfers: { select: { amount: true } },
       },
     })
-
-    //TODO: remove this when Prisma starts supporting nested groupbys
-    return campaigns.map(this.addReachedAmountAndDonors)
```
igoychev, line 82: We should not remove this, because we still need to know what was the reached amount on frontend.
In general, there are two aggregated amounts:
 - reachedAmount: this is sum(donations)
 - availableAmount: this is sum(donations)+sum(incomingTransfers)-sum(outgoingTransfers)-sum(expenses)-sum(withdrawals)

borislavstoychev: - reachedAmount: [link](https://github.com/podkrepi-bg/frontend/blob/356378b82e1fa60f78ada6afd0072fe51b2b9fc7/src/components/campaigns/grid/CampaignGrid.tsx#L54)
- availableAmount: [link](https://github.com/podkrepi-bg/frontend/blob/356378b82e1fa60f78ada6afd0072fe51b2b9fc7/src/components/campaigns/grid/CampaignGrid.tsx#L63-L69)
I did it on frontend because the function addReachedAmountAndDonors return empty vaults list. If we use this function blockedAmount can not be touched.

igoychev: ahaa, I see now, so addReachedAmountAndDonors also removes the vaults which is not said in the name ;)
then, to solve the problem, please move the removal of the vaults out of this function, so that you have better control on where to return them

stann1: I have a working solution that queries the db for all required sums and puts them in the summary prop. this branch ([enhancement/campaign-summaries](https://github.com/podkrepi-bg/api/tree/enhancement/campaign-summaries)

So I suggest to drop this change, because it still leads to loading of thousands of data points in a single request.

borislavstoychev: Great, i will close this PR.

borislavstoychev:
A better solution was implemented!

PR #299, opened 2022-07-18
 User Profile donation summary displayed only successful donations

borislavstoychev:
Requested changes from [frontend](https://github.com/podkrepi-bg/frontend/issues/893)

### DIFF
--- a/apps/api/src/donations/donations.service.ts
+++ b/apps/api/src/donations/donations.service.ts
@@ -325,11 +325,13 @@ export class DonationsService {
           include: { campaign: { select: { title: true } } },
         },
       },
-      where: { person: { keycloakId }, status: DonationStatus.succeeded },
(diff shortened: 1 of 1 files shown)


apps/api/src/donations/donations.service.ts
```diff
@@ -325,14 +325,13 @@ export class DonationsService {
           include: { campaign: { select: { title: true } } },
         },
       },
-      where: { person: { keycloakId } },
+      where: { person: { keycloakId }, status: DonationStatus.succeeded },
```
igoychev, line 328: If we filter on backend, this will hide the donations with statuses canceled from the User Profile donations table. Could you make the filter on the frontend only for the summary?

borislavstoychev:
Filtering on backend

PR #308, opened 2022-07-22
Feature/donations paging

stann1:


apps/api/src/app/middleware/apilogger.middleware.ts
```diff
@@ -0,0 +1,30 @@
...
+
+      if (statusCode >= 500) {
+        return this.logger.error(message)
+      }
+
+      if (statusCode >= 400) {
+        return this.logger.warn(message)
+      }
+
+      return this.logger.log(message)
```
igoychev, line 23: very useful addition of the logging middleware!
let's just not print every request but only the problematic ones - may be just comment out the logger.log(message) for those who want it, but no need on production.

apps/api/src/donations/donations.service.ts
```diff
@@ -157,23 +158,16 @@ export class DonationsService {
...
-      | 'extPaymentIntentId'
-      | 'extPaymentMethodId'
-      | 'billingName'
-      | 'billingEmail'
-    >[]
-  > {
-    return await this.prisma.donation.findMany({
+    pageIndex?: number,
+    pageSize?: number,
+  ): Promise<ListDonationsDto<DonationBaseDto>> {
```
igoychev, line 169: well done for extracting the return object as separate reusable dto!!!

apps/api/src/app/middleware/apilogger.middleware.ts
stann1: changed to debug level, so it does not show on prod

PR #342, opened 2022-10-15
New API endpoint to get all the campaigns that the user donated to

marinovl7:
I created this new API endpoint, because I'm implementing a new feature on the front end, that allows users to view the campaigns, that they donated to, that they are organizing, that they are coordinating or that they are the beneficiery of the campaign. This feature will allow the users later to also edit their camaigns and have everything sorted in a tab in MyProfile called My Campaigns, which will contain all the information about user's campaigns

I added a function in donations.service which retrieves the data from the database

I also added a function in the controller that defines the path that the API endpoint reffers to

The endpoint gets all the campaigns with all the data that the campaigns admin section have

### DIFF
--- a/apps/api/src/donations/donations.service.ts
+++ b/apps/api/src/donations/donations.service.ts
@@ -32,7 +32,7 @@ export class DonationsService {
     private campaignService: CampaignService,
     private prisma: PrismaService,
     private vaultService: VaultService,
-  ) { }
+  ) {}
 
   async listPrices(type?: Stripe.PriceListParams.Type, active?: boolean): Promise<Stripe.Price[]> {
     const list = await this.stripeClient.prices.list({ active, type })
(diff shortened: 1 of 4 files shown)


apps/api/src/donations/donations.service.ts
```diff
@@ -370,6 +371,41 @@ export class DonationsService {
     return { donations, total }
   }
 
+  async getUserDonatedCampaigns(keycloakId: string) {
+    const donations = await this.prisma.donation.findMany({
+      where: { person: { keycloakId } },
```
igoychev, line 376: The approach is good, yet after looking in more detail, we can do it in more optimal way.
To return the campaigns to which the user has donated, we can use the same method that lists all campaigns and add a filter by user. 

See campaign.service.ts listCampaigns() function and either create a new function or just extend this one to accept an optional AuthorizedUser parameter and then add this to the where clause: vaults: {some: { donations: { some: {person: {keycloakId}}}}}, 

Also please add this into the other branch so that the related changes stay together.

igoychev (approved):

PR #346, opened 2022-10-24
Edit permision to be given also for user in role, not only admins

### LINKED ISSUE #344: [Roles] Campaign Coordinator, Organizer and Beneficiary to be able to Edit campaign even if not in Admin Role
## Problem statement
Currently only users in Admin role can edit existing campaign. This forces us to give admin role to users that should be editing only "their" campaigns instead of all campaigns.

## Desired behavior
Extend the campaign PATCH endpoint to allow authenticated users who are Coordinator, Organizer or Beneficiary of the campaign to submit changes. To make this work pass @AuthenticatedUser() user: KeycloakTokenParsed parameter to the method and add the necessary checks. For reference, see [create-campaign] endpoint in campaign.controller.ts.

## Additional requirements
Please add necessary positive and negative tests to ensure this logic won't break with future changes

marinovl7:
I reimplemented the PATCH endpoint in campaign.controller to also allow user's, who are coordinator,beneficiary or organizer to edit their campaigns. This was requiered because of the front-end feature that displays the user's campaigns in my-campaigns tab in user's profile.
Closes #344


apps/api/src/campaign/campaign.controller.ts
```diff
@@ -119,8 +119,15 @@ export class CampaignController {
     roles: [RealmViewSupporters.role, ViewSupporters.role],
     mode: RoleMatchingMode.ANY,
   })
-  async update(@Param('id') id: string, @Body() updateCampaignDto: UpdateCampaignDto) {
-    return this.campaignService.update(id, updateCampaignDto)
+  async update(@Param('id') id: string, @Body() updateCampaignDto: UpdateCampaignDto, @AuthenticatedUser() user: KeycloakTokenParsed) {
+    const campaign = await this.campaignService.getCampaignByIdWithPersonIds(id)
+    if (user.sub === campaign?.beneficiary.person?.keycloakId ||
```
igoychev, line 124: since this is critical functionality for permisions check, please add tests to ensure all conditions are covered, so that we're protected from breaking the logic in future

marinovl7: I totally forgot about the tests.I'll implement the tests later tommorow

igoychev (approved):

Issue #359, opened 2022-11-10
[Improvement] Using external database keycloak


dimitur2204:
Exactly what is stated here, but in the `api` repo: https://github.com/podkrepi-bg/infrastructure/issues/17

PR #360, opened 2022-11-10
359 - One postgres instance for both keycloak and general database

dimitur2204:
Solves #359 

* Added a script to create a `keycloak` database on the current `postgres` instance.
* Changed the `keycloak` to now connect to the newly created database

Made the changes that @imilchev proposed in an old issue https://github.com/podkrepi-bg/infrastructure/issues/17 in the infrastructure repo. 
I don't know if it is still relevant though or if he is talking specifically about the infrastructure as code that is being deployed . In any case if it is needed in the infrastructure repo the solution should be similar

Notes:
There is a `CREATE SCHEMA "public";` line in the `init.sql` file which I am unsure if is relevant in any way, but couldn't test it.

This is how the file structure in the pg-db container looks like:
image

Is that right considering we have 2 databases now @imilchev?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -81,7 +81,7 @@ services:
     restart: always
 
     volumes:
-      - keycloak-db-data:/var/lib/postgresql/data
+      - keycloak-db-data:/var/lib/keycloak
       - pg-db-data:/var/lib/postgresql/data
     ports:
       - '${DB_PORT?}:5432'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -81,6 +81,7 @@ services:
     restart: always
 
     volumes:
+      # Keycloak creates it's own /data and /conf folders
       - keycloak-db-data:/var/lib/keycloak
       - pg-db-data:/var/lib/postgresql/data
     ports:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -81,8 +81,8 @@ services:
     restart: always
 
     volumes:
-      - pg-db-data:/bitnami/postgresql
-
+      - keycloak-db-data:/var/lib/postgresql/data
+      - pg-db-data:/var/lib/postgresql/data
     ports:
       - '${DB_PORT?}:5432'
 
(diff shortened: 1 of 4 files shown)


igoychev:
nice and quick - thanks for helping Nizamov! 
I was just thinking how we are going to do the migration now, because if we merge this it will redirect to the new empty database. So we need a migration script of some kind, don't we? - @imilchev what do you think?

docker-compose.yml
```diff
@@ -108,14 +107,17 @@ services:
...
       - KEYCLOAK_IMPORT=/opt/jboss/keycloak/config/keycloak-webapp-realm.json
       - DEBUG=true
       - DEBUG_PORT='*:8787'
-      - DB_VENDOR=H2
-      - TZ=Europe/Sofia
+      - DB_VENDOR=POSTGRES
+      - DB_USER=keycloak
+      - DB_PASSWORD=keycloak
+      - DB_ADDR=${COMPOSE_PROJECT_NAME?}-pg-db
+      - DB_DATABASE=keycloak
```
kachar, line 120: Looks like we're not using env vars for these items

kachar (approved):
Awesome upgrade @dimitur2204 :rocket:

kachar:
>  I was just thinking how we are going to do the migration now, because if we merge this it will redirect to the new empty database. So we need a migration script of some kind, don't we? - @imilchev what do you think?

@igoychev I think this setup is only local via docker-compose, so the data exports will be for the testing data only, is that correct?

dimitur2204:
> >  I was just thinking how we are going to do the migration now, because if we merge this it will redirect to the new empty database. So we need a migration script of some kind, don't we? - @imilchev what do you think?
> 
> @igoychev I think this setup is only local via docker-compose, so the data exports will be for the testing data only, is that correct?

I think this is the case for the code in this PR.
 
From what I understand from @imilchev his intention was to also do that on production though so that we only have 1 postgres instance to manage, however I think that will require help from him and everyone responsible for production!

PR #362, opened 2022-11-17
1087 feature add ability to assign legal entitycompany as organizer

dimitur2204:
**Purpose**
Add `company_id` field on the campaign regarding [#1087 ](https://github.com/podkrepi-bg/frontend/issues/1087) with a relation to the company field.

**Notes**
I noticed that sometimes we have these Relation objects and sometimes we don't
Here we have it:
image
And on this PR for example for the `campaign` we instead have a `toEntity()` function that seems to connect the fields:
image

Is there a difference?

### DIFF
--- a/.env
+++ b/.env
@@ -46,7 +46,7 @@ S3_SECRET_ACCESS_KEY=s3-secret-access-key
 
 ## Keycloak ##
 ##############
-KEYCLOAK_URL=http://localhost:8080/auth
+KEYCLOAK_URL=http://localhost:8180/auth
 KEYCLOAK_REALM=webapp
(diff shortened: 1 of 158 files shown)


kachar (approved):
Once again an awesome contribution @dimitur2204 

I can't see anything that can be improved around the code.

In your question - we're using `toEntity` so we can encapsulate which field are used to connect to external entity in the db. We validate the input uuids and link them annoyingly to reduce the duplication and reuse the code. The relation objects are a pattern that we couldn't validate enough to make sure we're consistent with it everywhere. The other approach seems to be more simplistic for the time being as the frontend doesn't have to know the proper json to match the Prisma linking style.

PR #367, opened 2022-11-24
366 Create a api-logs file containing the logs

dimitur2204:
Partly solves #366 
Added a` PodkrepiLogger`, extending the NestJS logger to now write to a file logs of level higher than LOG in addition to logging in the conosle. 
That was suggested by @imilchev since we couldn't debug a backend error while running in GitHub actions.

**However** I found that errors such as database connections or some unexpected crashes are not passed through the Logger which defeats the point, especially considering most of the errors in GitHub Actions are probably of that type! I still haven't found a solution and would appreciate help.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@
 /connect.lock
 /coverage
 /libpeerconnection.log
+api-error.log
 npm-debug.log
 yarn-error.log
 testem.log
--- a/.gitignore
+++ b/.gitignore
@@ -30,7 +30,7 @@
 /connect.lock
 /coverage
 /libpeerconnection.log
-api-error.log
+/api-logs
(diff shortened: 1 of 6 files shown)


dimitur2204:
[366 - add prismaClientInitialization errors to the catch decorator](https://github.com/podkrepi-bg/api/pull/367/commits/4fd07bc07aa3295fdd13959658eb5349bef6b718) This commit also doesn't seem to solve the problem of catching the base init error

imilchev:
I think we can close this one since we tackled the problem differently in https://github.com/podkrepi-bg/frontend/pull/1110

kachar:
Solved the file logging by running `yarn start &> api.log &` in the workflow

Issue #365, opened 2022-11-21
[dev] upgrade yarn to latest


igoychev:
**Is your feature request related to a problem? Please describe.**
We are using yarn v1 while latest is v3.3

**Describe the solution you'd like**
Looking at the upgrade instructions here: https://yarnpkg.com/getting-started/install it just requires running 
```
yarn set version stable
```
and then updating the .gitignore file as per https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

PR #372, opened 2022-11-29
upgrade yarn

quantum-grit:
Upgrading yarn to latest v3.3 to close https://github.com/podkrepi-bg/api/issues/365

Some side effects:
 - Upgrading yarn to v3.3 revealed a problem with jest versions and broke the tests
 - To fix the tests, I needed to upgrade jest packages to latest. This caused some config warnings, but tests pass.
 - Result: tests are now a lot faster passing for ~17secs! 

Note:  attempted to fix the warnings but that slows tests up to 5 times from 17sec to 125sec.
Todo: replace ts-jest with swc/jest for even fater tests: https://swc.rs/docs/usage/jest


Dockerfile
```diff
@@ -7,9 +7,11 @@ ENV TARGET_APP $TARGET_APP
 # Build target dependencies #
 #############################
 FROM base AS dependencies
-COPY package.json yarn.lock .yarnrc ./
+# Yarn
+RUN yarn set version berry
+COPY package.json yarn.lock .yarnrc.yml ./
 COPY schema.prisma .
-RUN yarn --production
+RUN yarn workspaces focus --all --production
```
kachar, line 14: We can copy `.yarn` in order to send the initial cache to the docker image

Not super important but prices becomes faster.

kachar (approved):

Dockerfile
igoychev: Yep, added now!

PR #375, opened 2022-11-30
fix migrations docker build

quantum-grit:
Upgrading yarn broke the migration docker build which is not tested in branch. Fixing now.

- fixed image after yarn upgrade
- add build-migrations to check-pr for testing in branch, will be removed before merging in master


.github/workflows/check-pr.yml
```diff
@@ -51,6 +51,81 @@ jobs:
         with:
           image: ghcr.io/podkrepi-bg/api/migrations:pr
 
+  build-migrations-docker-image:
+    name: Build migrations docker image
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Login to GitHub Packages Docker Registry
```
imilchev, line 60: this shouldn't be here. We are not pushing images from PRs since this is potentially unreviewed code

.github/workflows/check-pr.yml
```diff
@@ -51,6 +51,81 @@ jobs:
...
+
+      - name: Get sha
+        if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
+        run: echo "VERSION=master-$(echo $GITHUB_SHA | cut -c 1-6)" >> $GITHUB_ENV
+
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        env:
+          NODE_ENV: production
+        with:
```
imilchev, line 79: actually the migrations container is built in the previous step. Check line 38

.github/workflows/check-pr.yml
```diff
@@ -51,6 +51,81 @@ jobs:
...
+            ghcr.io/podkrepi-bg/api/migrations:${{ env.VERSION }}
+
+      - name: Scan with Mondoo
+        uses: mondoohq/actions/docker-image@main
+        env:
+          MONDOO_CONFIG_BASE64: ${{ secrets.MONDOO_SECRET }}
+        with:
+          image: ghcr.io/podkrepi-bg/api/migrations:${{ env.VERSION }}
+
+  build-seed-docker-image:
```
imilchev, line 96: I guess this is the only thing that needs to be added but you should definitely remove the login and push. You need the same steps as the build for the migrations container. The only difference should be the target of the build

imilchev (changes requested):

.github/workflows/check-pr.yml
quantum-grit: this was only temporary from the release.yml for testing the build, removed now. If anything needs to be fixed it needs to be updated in release.yml

.github/workflows/check-pr.yml
quantum-grit: this was only temporary from the release.yml for testing the build, removed now. If anything needs to be fixed it needs to be updated in release.yml

imilchev (approved):

Issue #368, opened 2022-11-24
Backend for view/upload reports to Campaigns


batebobo:
Backend of [[Feature] Add ability to view/upload reports to Campaigns](https://github.com/podkrepi-bg/frontend/issues/1093)
- [ ] DB schema changes to create reports and associate them to campaigns
- [ ] Endpoints and services for CRUD operations on reports

igoychev:
closed with the latest expenses implementation

PR #378, opened 2022-12-02
Backend for campaign reports

batebobo:
# Summary
Implements #368

Created CRUD endpoints for campaign reports. They are part of a separate module called `campaign-report`, but the controller is under the `campaign` base route.

New entities:
* CampaignReport
* CampaignReportFile

Both entities support soft delete and have an owner (uploader).

Screenshot of how the new routes look like:

Screenshot 2022-12-02 at 4 35 58 PM

## Todo
- [x] Endpoint for downloading files from reports
- [x] Tests for update service
- [x] Tests for update controller
- [x] Tests for delete service
- [x] Tests for delete controller
- [ ] Some manual testing of the whole process
- [ ] Manual test of s3 upload/download
- [ ] Squash migrations


schema.prisma
```diff
@@ -522,6 +525,48 @@ model BankTransactionsFile {
...
+  id                  String               @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
+  campaignId          String               @map("campaign_id") @db.Uuid
+  campaign            Campaign             @relation(fields: [campaignId], references: [id])
+  startDate           DateTime             @map("start_date")
+  endDate             DateTime             @map("end_date")
+  description         String
+  files               CampaignReportFile[]
+  totalFunds          Float?               @map("total_funds")
+  fundsForPeriod      Float?               @map("funds_for_period")
+  spentFundsForPeriod Float?               @map("spent_funds_for_period")
```
kachar, line 538: @batebobo Normally we use only `Integer` types for storing the amounts, so I would suggest that we keep that practice here as well to be consistent with the other use cases.

Stripe also provides the amounts in the smallest currency value.

Since the columns are optional we might add a default value of `0` so we always have a value.

```suggestion
  totalFunds          Int?              @map("total_funds") @default(0)
  fundsForPeriod      Int?              @map("funds_for_period") @default(0)
  spentFundsForPeriod Int?              @map("spent_funds_for_period") @default(0)
```

kachar (changes requested):
@batebobo Whooooah, incredible increment :rocket: :moon:

Kudos for the time spent on this!

I've noticed some incompatibilities with our current workflow regarding the storage of currency amounts. Please check them out and let's discuss.

Dockerfile.migrations
```diff
@@ -1,4 +1,5 @@
 FROM node:16-alpine3.16 as base
+RUN apk add g++ make py3-pip
```
kachar, line 2: @batebobo Let's add a comment on why we need these packages. If we need them how does it work without them now?

Please sync your branch with master as the node version has been bumped from 16 to 18 there.

migrations/20221128133021_add_campaign_reports/migration.sql
```diff
@@ -0,0 +1,43 @@
+-- CreateEnum
+CREATE TYPE "campaign_report_file_type" AS ENUM ('photo', 'document');
+
+-- CreateTable
+CREATE TABLE "campaign_reports" (
```
imilchev, line 5: Creating new tables breaks our database backup job currently due to permission changes. Please whoever merges this PR, ping me in a private message on Discord when it's merged. I will need to adjust the permissions before this is shipped to prod

apps/api/src/campaign-report/campaign-report.service.ts
```diff
@@ -0,0 +1,209 @@
...
+        id: reportId
+      },
+      data: {
+        isDeleted: { set: true }
+      }
+    })
+  }
+
+  private async softDeleteFile(fileId: string): Promise<CampaignReportFile> {
+    return this.prisma.campaignReportFile.update({
```
imilchev, line 129: when do we actually delete the files?

schema.prisma
batebobo: Yeah, makes sense, thanks :) I'll update the schema :v:

Dockerfile.migrations
batebobo: I forgot to ask a question on Discord about this. Without these I get an error that I need a python installation for `node-gyp` when running `docker-compose up`. I'll post a screenshot once I have a reliable internet connection to run a new `docker-compose up`

migrations/20221128133021_add_campaign_reports/migration.sql
batebobo: Thanks for letting us know! Is there I can do on this PR to avoid breaking the backup job? Can you also please post a reference to this job, so that I can understand better what it does and why creating tables is breaking it :pray:

apps/api/src/campaign-report/campaign-report.service.ts
batebobo: We don't for the moment as per [this reply](https://github.com/podkrepi-bg/frontend/issues/1093#issuecomment-1319845075) from @igoychev. We could implement the actual deletion in several different ways, but I'd propose to leave this for a future task if it's not absolutely required for the initial version of the feature.

Dockerfile.migrations
batebobo: Screenshot 2022-12-03 at 1 34 40 PM
This ^

I'll try updating the node version and see if it helps 😄

batebobo: It somehow works now 😩 Thank you!

kachar:
@imilchev The PR check for scanning k8s manifests with mondo showed an error that you might be interested in:

```shell
Target:     K8s Manifest dev-manifests
✕ Errors:   rpc error: code = InvalidArgument desc = asset does not match any of the activated policies
```

apps/api/src/campaign-report/campaign-report.controller.ts
```diff
@@ -0,0 +1,211 @@
...
+    const authorizedPerson = await this.userCanPerformReportAction(campaignId, user)
+    if (!authorizedPerson) {
+      throw new ForbiddenException('The user cannot modify the requested campaign')
+    }
+
+    const campaignReports = await this.campaignReportService.getReports(campaignId)
+
+    if (!campaignReports.map((report) => report.id).includes(reportId)) {
+      throw new NotFoundException('The given report is not part of the selected campaign')
+    }
```
igoychev, line 188: it will be more optimal instead of getting the reports to add a where clause in the database query

apps/api/src/campaign-report/campaign-report.controller.ts
```diff
@@ -0,0 +1,211 @@
...
+  private async userCanPerformReportAction(
+    campaignId: string,
+    user: KeycloakTokenParsed,
+  ): Promise<Person | null> {
+    const campaign = await this.campaignService.getCampaignByIdWithPersonIds(campaignId)
+    const userCanUploadReport = [
+      campaign?.beneficiary.person?.keycloakId,
+      campaign?.organizer?.person.keycloakId,
+      campaign?.coordinator.person.keycloakId,
+    ].includes(user.sub)
```
igoychev, line 202: more optimal will be to check in db again, see how it is done in getUserCampaigns(). It can be extended with an optional campaignId parameter and the query becomes:

this.prisma.campaign.findMany({
      where: {
        id: campaignId, 
        OR: [
          { beneficiary: { person: { keycloakId } } },
          { coordinator: { person: { keycloakId } } },
          { organizer: { person: { keycloakId } } },
        ],
      },

apps/api/src/domain/generated/person/entities/person.entity.ts
```diff
@@ -36,6 +38,8 @@ export class Person {
   benefactors?: Benefactor[]
   beneficiaries?: Beneficiary[]
   campaignFiles?: CampaignFile[]
+  campaignReports?: CampaignReport[]
+  campaignReportFiles?: CampaignReportFile[]
```
igoychev, line 42: shouldn't the relation be from person to campaignReport and then from campaignReport to the files? 
In other words the Person should not have direct relation to the Files

schema.prisma
```diff
@@ -522,6 +525,48 @@ model BankTransactionsFile {
...
+  description         String
+  files               CampaignReportFile[]
+  totalFunds          Int?               @map("total_funds") @default(0)
+  fundsForPeriod      Int?               @map("funds_for_period") @default(0)
+  spentFundsForPeriod Int?               @map("spent_funds_for_period") @default(0)
+  goals               String
+  nextSteps           String               @map("next_steps")
+  additionalInfo      String               @map("additional_info")
+  creatorId           String               @map("creator_id") @db.Uuid
+  creator             Person               @relation(fields: [creatorId], references: [id])
```
igoychev, line 543: let's unify the naming convention:
creator could be createdBy or uploadedBy

This will allow in the next iterations to add columns for 'approvedBy' or 'deletedBy'. Similarly we have the need to add some timestamps like 'createdAt, deletedAt to record when such actions happened

schema.prisma
```diff
@@ -522,6 +525,48 @@ model BankTransactionsFile {
...
+
+model CampaignReportFile {
+  id        String                 @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
+  reportId  String                 @map("report_id") @db.Uuid
+  report    CampaignReport         @relation(fields: [reportId], references: [id])
+  filename  String                 @db.VarChar(200)
+  mimetype  String                 @db.VarChar(50)
+  type      CampaignReportFileType
+  personId  String                 @map("person_id") @db.Uuid
+  person    Person                 @relation(fields: [personId], references: [id])
```
igoychev, line 564: [naming convention] let's have these uploadedBy to indicate that they indicate who uploaded them. In other context someone may confuse the file is "about the person".

igoychev (commented):
Great start and happy to see this code starts to take the shape! 
I added some comments to help you with Prisma and naming conventions

apps/api/src/campaign-report/campaign-report.controller.ts
```diff
@@ -0,0 +1,211 @@
...
+  @Public()
+  async getReport(
+    @Param('campaignId') campaignId: string,
+    @Param('reportId') reportId: string,
+  ): Promise<GetReportDto> {
+    const campaignReports = await this.campaignReportService.getReports(campaignId)
+
+    if (!campaignReports.map((report) => report.id).includes(reportId)) {
+      throw new NotFoundException('The given report is not part of the selected campaign')
+    }
```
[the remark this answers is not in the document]
batebobo: `getReports` and `getReport` have different return types. In `getReports` we want the basic information about a report without fetching any of the relations. In `getReport` we'd like a more complicated fetch. Does that clarify the need for the two service functions?

migrations/20221128133021_add_campaign_reports/migration.sql
```diff
@@ -0,0 +1,43 @@
...
+    "mimetype" VARCHAR(50) NOT NULL,
+    "type" "campaign_report_file_type" NOT NULL,
+    "person_id" UUID NOT NULL,
+
+    CONSTRAINT "campaign_reports_files_pkey" PRIMARY KEY ("id")
+);
+
+-- AddForeignKey
+ALTER TABLE "campaigns" ADD CONSTRAINT "campaigns_company_id_fkey" FOREIGN KEY ("company_id") REFERENCES "companies"("id") ON DELETE SET NULL ON UPDATE CASCADE;
+
```
igoychev, line 35: this was done also on master and now having it two times breaks the migrations

kachar:
@batebobo Can you please update your PR and resolve the open discussions in order to get this PR moving? :pray:

igoychev:
closing in favor of https://github.com/podkrepi-bg/api/pull/453

Issue #379, opened 2022-12-02
[Improvement] Add better database seeds


dimitur2204:
We had a problem that broke production which we didn't catch neither from the backend, nor the frontend testing, due to insufficient and not true to life database seeds.

The exact case was that the `beneficiary` can be a company and not only a person. And on the frontend we had a call assuming `person` always exists, and the e2e tests didn't catch that since we didn't have such a case on the seeds.

They are one of the places where we can improve!

batebobo:
It would be best if you could describe what exactly broke production and how you'd like to improve the seeds. Right now the task is too generic and It's not clear (to me) what the desired outcome is 🙏

dimitur2204:
> The exact case was that the `beneficiary` can be a company and not only a person. And on the frontend we had a call assuming `person` always exists, and the e2e tests didn't catch that since we didn't have such a case on the seeds.

@batebobo

PR #380, opened 2022-12-03
379 Add better database seeds

dimitur2204:
Added the companies to the seeds, because we broke production last time because we did not have a case where a beneficiary is a company in the seeds.
#379 
* Upgraded `faker` to the new `@faker-js/faker`
* Added seeds for companies
* Added beneficiaries as companies
* Added company beneficiaries to campaigns


kachar (approved):
Nice one @dimitur2204 

I wonder if we can do some tricks to improve the DX in the seeds. Right now we insert static data in the db in one file and then we fetch given records in another file. I guess we can statically export and import the dummy data instead of going to the db and back.

This will make our seeds independent from the db state.

That's only food for thoughts 💭

igoychev (approved):
nice improvement for increased safety!

Issue #374, opened 2022-11-29
Add foreign key for campaign.company_id


igoychev:
In the current database schema in schema.prisma file the Campaign table contains companyId field, which is a relation to Company table. However, the migrations don't contain the creation of the foreighnKey in the database.

To add the foreign key, you have to execute:
`yarn prisma migrate dev`
and then follow the instructions of the wizard to create the new migration that will add the foreign key.

PR #383, opened 2022-12-04
Add foreign key for campaign.company_id

Nnachevvv:
Implements #374
After execution, the campaign table contains `campaigns_company_id_fkey` as constraint:
```
 constraint_catalog | constraint_schema |             constraint_name             | table_catalog | table_schema |       table_name        | constraint_type | is_deferrable | initially_deferred | enforced
 postgres           | api               | campaigns_company_id_fkey               | postgres      | api          | campaigns               | FOREIGN KEY     | NO            | NO                 | YES
```

### DIFF
--- a/migrations/20221204140556_add_company_id_foreign_key_for_campaign/migration.sql
+++ b/migrations/20221204140556_add_company_id_foreign_key_for_campaign/migration.sql
@@ -1,0 +1,2 @@
+-- AddForeignKey
+ALTER TABLE "campaigns" ADD CONSTRAINT "campaigns_company_id_fkey" FOREIGN KEY ("company_id") REFERENCES "companies"("id") ON DELETE SET NULL ON UPDATE CASCADE;


igoychev (approved):
quick and tidy! thank you!

kachar:
@Nnachevvv Was there a need to change `schema.prisma` file or the migration was enough? I don't see such diff in the PR

Nnachevvv:
@kachar I think this field has been already added in our schema before:
https://github.com/podkrepi-bg/api/blob/d75283f2b172ec61c1845e534cec78016df24479/schema.prisma#L203


The command `yarn prisma db pull` only aligns the lines, so I think the migration is enough, but I maybe be wrong

kachar:
Ah, so the field has been added but without a migration, and now we add the migration. Thanks for clarifying that and kudos for your efforts. 🚀

PR #384, opened 2022-12-05
fixed rounding when importing bank transactions with floating point

quantum-grit:
Error was reported when importing bank transaction file for a specific campaign as per: https://github.com/podkrepi-bg/frontend/issues/1136

It turned out the problem is not that the campaign is completed, but because of a single amount of 160.20 which was parsed as 160.1999999. The fix was to add rounding for the expected amount to be imported.

### DIFF
--- a/apps/api/src/bank-transactions-file/helpers/parser.ts
+++ b/apps/api/src/bank-transactions-file/helpers/parser.ts
@@ -19,7 +19,13 @@ export function parseBankTransactionsFile(
         ) {
           const payment = new CreateManyBankPaymentsDto()
           const paymentRef = items[item].AccountMovement[movement].NarrativeI02[0]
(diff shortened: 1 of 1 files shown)


kachar (commented):
@quantum-grit I guest that we extract this functionality as utility method as we are gonna face the same scenario in the future as well.


```tsx
function normalizeAmount(amount: number | string): number {
  const value = Number(amount) // or parseFloat()
  if (isNaN(value)) return 0
  return Math.round((value + Number.EPSILON) * 100 * 100) / 100
}
```

This will help us test the conversion with the unit tests

dimitur2204 (approved):
@kachar's comment seems fair!

quantum-grit:
New PR submitted due to forking limitations: https://github.com/podkrepi-bg/api/pull/385

PR #386, opened 2022-12-06
added the missing campaignPhotos in response

quantum-grit:
Due to optimizations the campaignList endpoint was not returning the campaignListPhotos.
Fixed by adding them in response.


kachar (approved):
:medal_sports:

PR #387, opened 2022-12-06
Fix date time for bank transactions

Nnachevvv:
Fixes: https://github.com/podkrepi-bg/frontend/issues/1204

The fix for this is to add the wanted timezone(GMT+0200) when we parse the PaymentDateTime record.


apps/api/src/bank-transactions-file/helpers/parser.ts
```diff
@@ -6,6 +6,7 @@ import { toMoney } from '../../common/money'
 export const parseString = require('xml2js').parseString
 
 const regexPaymentRef = /\b[A-Z0-9]{4}-[A-Z0-9]{4}-[A-Z0-9]{4}\b/g
+const timeZoneGMTEasternEurope = '+0200'
```
igoychev, line 9: Well done, however just thought that we need to adjust for the Daylight Saving Time which is +3 in summer.

As per [Wikipedia for Time in Bulgaria](https://en.wikipedia.org/wiki/Time_in_Bulgaria): 
Daylight saving time moves one hour ahead to [UTC+03:00](https://en.wikipedia.org/wiki/UTC%2B03:00) is observed from the last Sunday in March to the last Sunday in October,[[2]](https://en.wikipedia.org/wiki/Time_in_Bulgaria#cite_note-2) inline with most [EU](https://en.wikipedia.org/wiki/EU) member states.[[3]](https://en.wikipedia.org/wiki/Time_in_Bulgaria#cite_note-3)

kachar: We can use a standard approach to time conversion 

```tsx
const now = new Date();
const sofiaTime = now.toLocaleString("bg-BG", {timeZone: "Europe/Sofia"});
console.log(sofiaTime); // Output: "12/8/2022, 9:31:03 PM"
```

Nnachevvv: Good catch! Thanks for your suggestions. I decided to use Intl.DateTimeFormat with a slice to get the TimezoneOffset. Unfortunately, I couldn't find an easier and cleaner way to do this

kachar (approved):
Awesome upgrade @Nnachevvv 
Thanks for the dedication and time spent on this :calendar: :clock1030:

Nnachevvv:
I saw that tests failed with the following error:
TS2322: Type '"longOffset"' is not assignable to type '"long" | "short" | undefined'.
[67](https://github.com/podkrepi-bg/api/actions/runs/3658209518/jobs/6182703744#step:8:68)
I will take a look why this has happened.

Nnachevvv:
Please still don't merge this. After the update https://github.com/podkrepi-bg/api/pull/392 lint is fixed and option longOffset will become available, but now after testing the change it fails with:
`Value longOffset out of range for Intl.DateTimeFormat options property timeZoneName'
Which I guess it is coming from the browser/OS. I'm still investigating this and perhaps we should change the approach and use another library.

Nnachevvv:
I have found a better approach. I have added [luxon library](https://moment.github.io/luxon/#/?id=luxon), which seems to have better browser and DST support.

Testing with following file
[22_11_18_TestImport.xml.txt](https://github.com/podkrepi-bg/api/files/10201879/22_11_18_TestImport.xml.txt)
Paymant dates defined in file:
```
    <PaymentDateTime>2022-11-18T09:03:31</PaymentDateTime>
    <PaymentDateTime>2022-11-17T17:57:11</PaymentDateTime>
    <PaymentDateTime>2022-11-17T16:49:15</PaymentDateTime>
```
Result after import:

Screenshot 2022-12-11 at 12 14 02

kachar:
https://github.com/podkrepi-bg/api/pull/392 has been merged and TS is upgraded now. This task should be unblocked

Nnachevvv:
@kachar Yes, I think it is okay to merge this change.

igoychev (approved):
well done!

PR #390, opened 2022-12-07
Recurring donations backend part

slavcho:
1. Add support for subscriptions in the stripe calls.
2. Listen on the create and update subscription event hooks.
3. Listen on the invoice paid hook in order to add the actual payment. For restrictions we don't have payment intents.
4. Make sure that when the campaign is completed - we cancel all active subscriptions.

Issue:
https://github.com/podkrepi-bg/frontend/issues/1088


apps/api/src/campaign/campaign.service.ts
```diff
@@ -140,6 +140,34 @@ export class CampaignService {
...
+
+
+    if (campaign.vaults.length == 0) {
+      throw new NotFoundException('No vaults found for campaign: ' + campaignId)
+    }
+
+    for(const vault of campaign.vaults) {
+      campaign['defaultVault'] = vault.id
+      break
+    }
```
igoychev, line 165: There is something strange here. If we just want to assign the first vault as default, then no need of a for cycle at all.

slavcho: I agree. I will fix it.

igoychev (commented):
Looks like a great addition! Well done with finding the right places to plugin with the recurring donations!

apps/api/src/donations/donations.service.ts
```diff
@@ -115,38 +115,67 @@ export class DonationsService {
...
     })
 
-    await this.createInitialDonation(campaign, sessionDto, session.payment_intent as string)
+    if (mode == 'payment') {
+      await this.createInitialDonation(
+        campaign,
+        sessionDto,
+        (session.payment_intent as string) ?? session.id,
+      )
+    }
```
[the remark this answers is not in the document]
slavcho: Sure. I will add this.

apps/api/src/donations/events/stripe-payment.service.ts
```diff
@@ -110,6 +114,152 @@ export class StripePaymentService {
...
+      throw new BadRequestException('Subscription does not contain any items. Subscription is: ' + subscription)
+    }
+
+    const priceItem = subscription.items.data[0]
+
+    if (priceItem.price.unit_amount) {
+      rdDto.amount = priceItem.price.unit_amount as number
+    } else if (priceItem.plan.billing_scheme == 'tiered' && priceItem.quantity) {
+      //for tiered plans - we have price of 1 BGN and the quantity matches the amount
+      rdDto.amount = priceItem.quantity * 100
```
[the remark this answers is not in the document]
slavcho: Nope. Good catch.
Removed it.

apps/api/src/recurring-donation/recurring-donation.controller.ts
```diff
@@ -44,6 +57,25 @@ export class RecurringDonationController {
     return this.recurringDonationService.update(id, updateRecurringDonationDto)
   }
 
+  @Get('cancel/:id')
+  @Roles({
+    roles: [RealmViewSupporters.role, ViewSupporters.role],
+    mode: RoleMatchingMode.ANY,
+  })
+  cancel(@Param('id') id: string) {
```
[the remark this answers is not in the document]
slavcho: Yep... missed that. 
It is fixed now

apps/api/src/donations/events/stripe-payment.service.ts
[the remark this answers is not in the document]
slavcho: fixed.

apps/api/src/recurring-donation/recurring-donation.controller.ts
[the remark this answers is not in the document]
slavcho: Done.

apps/api/src/donations/donations.service.ts
```diff
@@ -120,38 +120,65 @@ export class DonationsService {
...
+
+    Logger.log('[ CreateInitialDonation ]', {
+      session: session,
     })
 
-    await this.createInitialDonation(campaign, sessionDto, session.payment_intent as string)
+    await this.createInitialDonation(
+      campaign,
+      sessionDto,
+      (session.payment_intent as string) ?? session.id,
```
kachar, line 153: I'm not sure the default `session.id` should be added here instead of `paymentIntentId` when `session.payment_intent` is not present.

This might lead to storing different data in the same db column.

apps/api/src/donations/events/stripe-payment.service.ts
```diff
@@ -110,6 +114,157 @@ export class StripePaymentService {
...
+  }
+
+  @StripeWebhookHandler('customer.subscription.created')
+  async handleSubscriptionCreated(event: Stripe.Event) {
+    const subscription: Stripe.Subscription = event.data.object as Stripe.Subscription
+    const recurringDonation = await this.recurringDonationService.findSubscriptionByExtId(subscription.id)
+    if (recurringDonation) {
+      //we already have this subscription in our database
+      //stripe already sent us that event
+      return
```
kachar, line 129: Please at least add log for this scenario since it will be hard to track later on if we want to check why we're not handling this subscription event.

apps/api/src/donations/events/stripe-payment.service.ts
```diff
@@ -110,6 +114,157 @@ export class StripePaymentService {
...
+    const campaign = await this.campaignService.getCampaignByIdWithDefaultVault(metadata.campaignId)
+
+    const rdDto: CreateRecurringDonationDto = new CreateRecurringDonationDto()
+    rdDto.campaign = metadata.campaignId
+    rdDto.extSubscriptionId = subscription.id,
+    rdDto.extCustomerId = subscription.customer as string
+    rdDto.sourceVault = campaign["defaultVault"]
+
+    if (subscription.items.data.length == 0) {
+      throw new BadRequestException('Subscription does not contain any items. Subscription is: ' + subscription)
```
kachar, line 152: We cast object `subscription` to string here. Maybe we need to pass only the subscription id

apps/api/src/donations/events/stripe-payment.service.ts
```diff
@@ -110,6 +114,157 @@ export class StripePaymentService {
...
+          campaign.currency
+        } <> donation currency ${invoice.currency.toUpperCase()}`,
+      )
+    }
+
+    const billingData = getInvoiceData(invoice)
+
+    this.campaignService.donateToCampaign(campaign, billingData).then(() => {
+      this.checkForCompletedCampaign(metadata.campaignId as string)
+    })
```
kachar, line 257: I don't get why we use `then()` syntnax here instead of `await this.campaignService.donateToCampaign`

This breaks the synchronous flow in my opinion.

apps/api/src/recurring-donation/recurring-donation.service.ts
```diff
@@ -39,15 +144,60 @@ export class RecurringDonationService {
...
+
+      // the webhook will handle this as well.
+      // but we cancel it here, in case the webhook is slow.
+      this.findSubscriptionByExtId(result.id).then((rd) => {
+        if (rd) {
+          return this.cancel(rd.id)
+        }
+      })
+    })
+  }
```
kachar, line 202: Let's use `async/await` syntnax instead of `then()` promise callbacks

apps/api/src/recurring-donation/recurring-donation.service.ts
```diff
@@ -39,15 +144,60 @@ export class RecurringDonationService {
...
+    id: string,
+    status: RecurringDonationStatus,
+  ): Promise<RecurringDonation | null> {
+    const result = await this.prisma.recurringDonation.update({
+      where: { id: id },
+      data: {
+        status: status,
+      },
+    })
+    if (!result) throw new NotFoundException('Not found')
```
kachar, line 182: @slavcho This method should throw `BadRequestException` and not `NotFoundException` as it will send status code 404 to the client. 

We actually have a proper HTTP handler for the method, but the params that are sent to it are bad so we better respond with status code 400.

apps/api/src/recurring-donation/recurring-donation.service.ts
```diff
@@ -18,9 +29,103 @@ export class RecurringDonationService {
     return await this.prisma.recurringDonation.findMany()
   }
 
+  static defaultSelectFields = {
```
kachar, line 32: You can use `Prisma.validate()` to better type this field.

https://www.prisma.io/docs/concepts/components/prisma-client/advanced-type-safety/prisma-validator

You can also use `includes` key instead of `select` key to skip mentioning every single field in the initial model and only specify the relations like `sourceVault` and `person`_

apps/api/src/donations/donations.service.ts
slavcho: well.. I am open to suggestions...
Initially I have skipped creating "initial" donations for subscriptions.
@igoychev convinced me that it is better to have some track record that a payment has been attempted. 
When the actually payment is received it will be attached and the fields will be properly populated.

apps/api/src/donations/events/stripe-payment.service.ts
slavcho: Sure - will do.

apps/api/src/donations/events/stripe-payment.service.ts
slavcho: fixed.

apps/api/src/recurring-donation/recurring-donation.service.ts
slavcho: OK

apps/api/src/recurring-donation/recurring-donation.service.ts
slavcho: Fixed. Thanks.

apps/api/src/donations/donations.service.ts
igoychev: @kachar we already use the extPaymentIntentId to hold different data - for example for bank payments it holds the bank DocumentReferenceId. The safety comes from the extPaymentMethodId which indicates what type of reference this is - currently card or bank. 

So to make it safe for subscriptions too, one proposal is to set extPaymentMethodId to 'stripe-subscription'.

slavcho: We are already doing that:

        extPaymentMethodId: sessionDto.mode === 'subscription' ?  'subscription' : 'card',

kachar:
Awesome update @slavcho! Kudos for the time spent on that

apps/api/src/donations/donations.service.ts
slavcho: Sounds good. I can try to prepare a new PR for this.
Will check during the weekend.

kachar: Another thing that was brought to my attention is that `extPaymentMethodId` column expects the stripe expects an id like `pm_1MEzEO2eZvKYlo2CPlygE9a3` of the actual card that has done the payment. As I see we place hardcoded strings like `subscription` or `card` which is not correct in terms of data architecture.

apps/api/src/campaign/campaign.service.ts
```diff
@@ -140,6 +140,31 @@ export class CampaignService {
...
+      Logger.warn('No campaign record with ID: ' + campaignId)
+      throw new NotFoundException('No campaign record with ID: ' + campaignId)
+    }
+
+
+    if (campaign.vaults.length == 0) {
+      throw new NotFoundException('No vaults found for campaign: ' + campaignId)
+    }
+
+    campaign['defaultVault'] = campaign.vaults[0].id
```
kachar, line 162: When we alter the campaign object we don't actually update the response type of Prisma, so it won't be available during static typing checks.

If we decide to keep this approach we might rename `defaultVault` to `defaultVaultId` to better state the content we keep in.

I'd suggest that we have a helper method like:

```tsx
getDefaultVaultId(campaign: Campaign): string {
  if (campaign.vaults.length == 0) {
    throw new NotFoundException('No vaults found for campaign: ' + campaignId)
  }
  return campaign.vaults[0].id
}
```

apps/api/src/donations/donations.controller.ts
```diff
@@ -42,7 +43,20 @@ export class DonationsController {
 
   @Post('create-checkout-session')
   @Public()
-  createCheckoutSession(@Body() sessionDto: CreateSessionDto) {
+  async createCheckoutSession(@Body() sessionDto: CreateSessionDto) {
+    if (sessionDto.mode == 'subscription' && (sessionDto.personId == null || sessionDto.personId.length == 0)) {
```
kachar, line 47: It's a best practice to use strict comparison checks using `===` instead of `==`

Read more at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#strict_equality_using

```suggestion
    if (sessionDto.mode === 'subscription' && (sessionDto.personId === null || sessionDto.personId.length === 0)) {
```

apps/api/src/campaign/campaign.service.ts
slavcho: It turns out we already have: getCampaignVault.
I am using this now.

apps/api/src/donations/donations.controller.ts
slavcho: Thanks.
Only in Javascript you can have 2 == "2" as true.

apps/api/src/campaign/campaign.service.ts
kachar: That's even better! Code reuse for the win :top:

igoychev (approved):
well done overall, we're merging this and more improvements will be done in new PRs

apps/api/src/donations/events/stripe-payment.service.ts
```diff
@@ -110,6 +122,193 @@ export class StripePaymentService {
...
+      )
+    }
+
+    const vault = await this.campaignService.getCampaignVault(metadata.campaignId)
+    if (!vault) {
+      throw new BadRequestException('Vault for campaign ' + metadata.campaignId + ' does not exist')
+    }
+
+    const rdDto: CreateRecurringDonationDto = new CreateRecurringDonationDto()
+    rdDto.campaign = metadata.campaignId
```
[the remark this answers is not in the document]
slavcho: done

PR #391, opened 2022-12-09
Export data to excel

AndreyGoranov:
Issue: https://github.com/podkrepi-bg/frontend/issues/700

- Add Export module. 
- Export service to be used around the app in different services that would require exporting.
- There is an option to specify everything about the template in a single JSON or you could use a default setup with all columns from the data provided for exporting.
- Add new dependency: https://github.com/exceljs/exceljs
- Use types for the exportable data and add types for the template as well


kachar (commented):
@AndreyGoranov Everything looks smooth and nice!

The only thing that looks strange to me is the 15k rows diff by the added `yarn.lock` file. Looks like it's been changed entirely. Can you check that please?

AndreyGoranov:
@kachar yes will do I had a merge conflict on it and probably I didn't resolve it right

kachar:
@AndreyGoranov You can delete the local file and run `yarn` to regenerate it.

I hope you're using yarn 3 (`yarn set version berry`)

AndreyGoranov:
@kachar I think its ok now ?

apps/api/src/export/export.service.ts
```diff
@@ -0,0 +1,30 @@
+import { Injectable } from '@nestjs/common'
+import { PrismaService } from '../prisma/prisma.service'
+import { Response } from 'express'
+import { createWorkbook } from './helpers/createWorkbook'
+import { ExportableData } from '@prisma/client'
```
kachar, line 5: @AndreyGoranov Can you help me understand what is `ExportableData` as I can find an update of `schema.prisma` and this type is not being generated.

AndreyGoranov: It's just a Type:  export Type ExportableData = Donations[] | Campaigns[]

kachar (approved):
@AndreyGoranov Nice upgrade of the api!

It wasn't easy to make the tests and check green, but we've managed it. Thanks for the collaboration along the way

igoychev (commented):
Change requested: 
The amounts in the exported file should be divided by 100 to contain the cents.

kachar:
@AndreyGoranov Can you please create a PR with the update that @igoychev mentioned above?

AndreyGoranov:
https://github.com/podkrepi-bg/api/pull/396

PR #392, opened 2022-12-09
Upgrade typescript dependency from 4.3.5->4.8.3

Nnachevvv:
This PR makes version of typescript to be same as the version in the frontend repo. Also, the following [PR](https://github.com/podkrepi-bg/api/pull/387) will be unblocked, as it has been failing with a feature which is missing from the older typescript version.


kachar (approved):
Nicely done @Nnachevvv

PR #393, opened 2022-12-11
Check if billing name is present during cash deposit

Nnachevvv:
Implements 
https://github.com/podkrepi-bg/frontend/issues/1205

Test:

I've used following xml file:
[22_11_18_TestImport.xml.txt](https://github.com/podkrepi-bg/api/files/10201806/22_11_18_TestImport.xml.txt)

I have imported following records:
```xml
    <MovementFunctionalType>Вноска на каса</MovementFunctionalType>
    <MovementType>Credit</MovementType>
    <Narrative nil="true" />
    <OppositeSideAccount>BG78CECB123123123</OppositeSideAccount>
    <OppositeSideName nil="true" />
    <PaymentDate>2022-11-18T00:00:00</PaymentDate>
    <Reason>Получен междубанков превод</Reason>
------------------------------------------------------------------------------------------------
    <MovementFunctionalType>Вноска на каса</MovementFunctionalType>
    <MovementType>Credit</MovementType>
    <Narrative nil="true" />
    <OppositeSideAccount>BG11RZBB123123123123</OppositeSideAccount>
    <OppositeSideName>Firstname Lastnameee</OppositeSideName>
    <PaymentDate>2022-11-17T00:00:00</PaymentDate>
------------------------------------------------------------------------------------------------

    <CCY>BGN</CCY>
    <MovementFunctionalType>Получен междубанков превод</MovementFunctionalType>
    <MovementType>Credit</MovementType>
    <Narrative nil="true" />
    <OppositeSideAccount>BG12FINV9150123123123</OppositeSideAccount>
    <OppositeSideName>Първо Второ Трето Име</OppositeSideName>
    <PaymentDate>2022-11-17T00:00:00</PaymentDate>
```

Billing name is not present when  OppositeSideName nil="true"  and  MovementFunctionalType  is "Вноска на каса".:
Screenshot 2022-12-11 at 11 39 22

### DIFF
--- a/apps/api/src/bank-transactions-file/helpers/parser.ts
+++ b/apps/api/src/bank-transactions-file/helpers/parser.ts
@@ -35,7 +35,7 @@ export function parseBankTransactionsFile(
           const movementFunctionalType =
             items[item].AccountMovement[movement].MovementFunctionalType[0]
           const billingName = items[item].AccountMovement[movement].OppositeSideName[0]
-          if (isBillingNamePresent(movementFunctionalType, billingName)) {
+          if (!isBillingNameNil(movementFunctionalType, billingName)) {
             payment.billingName = billingName
           }
 
@@ -52,8 +52,17 @@ export function parseBankTransactionsFile(
   return accountMovements
 }
 
-function isBillingNamePresent(movementFunctionalType: string, billingName: string) {
-  return !(
+/**
+ * The function checks if the movementType is cash deposit and if billing name is present
+ * @param movementFunctionalType - transaction type
+ * @param billingName - name of the billing if present, otherwise a json object
+ * @returns boolean
+ **/
+function isBillingNameNil(
+  movementFunctionalType: string,
+  billingName: string | { [x: string]: unknown },
+): boolean {
+  return (
     movementFunctionalType === cashDepositBGString &&
     billingName['$'] !== undefined &&
     billingName['$']['nil']
--- a/apps/api/src/bank-transactions-file/helpers/parser.ts
+++ b/apps/api/src/bank-transactions-file/helpers/parser.ts
@@ -6,6 +6,7 @@ import { toMoney } from '../../common/money'
 export const parseString = require('xml2js').parseString
 
 const regexPaymentRef = /\b[A-Z0-9]{4}-[A-Z0-9]{4}-[A-Z0-9]{4}\b/g
+const cashDepositBGString = 'Вноска на каса'
 
 export function parseBankTransactionsFile(
   fileBuffer,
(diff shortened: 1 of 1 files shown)


apps/api/src/bank-transactions-file/helpers/parser.ts
```diff
@@ -39,3 +46,11 @@ export function parseBankTransactionsFile(
   })
   return accountMovements
 }
+
+function isBillingNamePresent(movementFunctionalType: string, billingName: string) {
+  return !(
```
kachar, line 51: It's really tricky to see the reverse sign `!` here

I suggest we take the `not` outside of the function

```tsx
if (!isBillingNamePresent()) {

}
```

We can rename the function to state its purpose better if we need to.

Some explanation as comments would be useful here too

apps/api/src/bank-transactions-file/helpers/parser.ts
```diff
@@ -39,3 +46,11 @@ export function parseBankTransactionsFile(
...
+    movementFunctionalType === cashDepositBGString &&
+    billingName['$'] !== undefined &&
```
kachar, line 53: It's a bit strange to expect `billingName` to be string in the fn definition and then to use it array here.

Please improve the types of the input.

kachar (changes requested):

apps/api/src/bank-transactions-file/helpers/parser.ts
Nnachevvv: This is a tricky one, because:
When we have   <OppositeSideName>Firstname Lastnameee</OppositeSideName> we pass string to the function.
On the other hand  <OppositeSideName nil="true" /> is passed as Json object.
I used the following option:
```
    string | { [x: string]: unknown }
```

I'm open to any suggestions, if there is a better way to improve that.

kachar: @Nnachevvv  The update looks great, brings a lot more context, and clears up the duality of the types.

Thanks for the contribution!

kachar (approved):

apps/api/src/bank-transactions-file/helpers/parser.ts
kachar: @Nnachevvv  After some more thinking about this piece of code i figured out the intent was not super clear as we have two separate functionality in the same function right now. We can separate them as:


```tsx
const oppositeSide = items[item].AccountMovement[movement].OppositeSideName[0]
if (isCashDeposit(movementFunctionalType) && !isBillingNameNil(oppositeSide)) {
  payment.billingName = oppositeSide
}

function isCashDeposit(movementFunctionalType: string): boolean {
  return movementFunctionalType === cashDepositBGString
}
function isBillingNameNil(
  billingName: string | { [x: string]: unknown },
): boolean {
  return (
    billingName['$'] !== undefined &&
    billingName['$']['nil']
  )
}
```

Nnachevvv: I agree with you. It looks better , I will make a new PR with the change. Thanks!

PR #396, opened 2022-12-12
Update the mapping of donations data before export

AndreyGoranov:


kachar (approved):
Thanks @AndreyGoranov

PR #397, opened 2022-12-13
Add pagination to findWishesByCampaignId endpoint.

Katzarov:
Back-end part for #[1129](https://github.com/podkrepi-bg/frontend/issues/1129)
Front-end PR https://github.com/podkrepi-bg/frontend/pull/1252

- implements offset pagination for findWishesByCampaignId endpoint.
- **this change is not compatible with currently deployed Frontend** (I assumed making things backwards compatible is not a concern, and such changes are "just" deployed together)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 
-FROM node:18-alpine3.16 as base
+FROM node:18-alpine3.17 as base
 WORKDIR /app
+RUN apk add --update --no-cache openssl1.1-compat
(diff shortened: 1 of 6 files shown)


kachar (approved):
Thanks for the contribution @Katzarov 

Works flawlessly :field_hockey: 

Added seeds for the donation wishes so every dev can work with actual data

kachar:
We'll fix the prisma build problem on master, and not on the PR

PR #403, opened 2022-12-18
Introduce factories for the seeds

valkirilov:
Introduce factories for the seeds

- integrated fishery which provides clean and elegant solution for creating model factories easily (for seeding, testing, etc.)
- created a PoC with the Person seeds to showcase the factories idea
- minor refactoring to the seeds scripts

_Note: This is just a proposal for an enhancement and a PoC applied to only one of the models. In case this approach is accepted, I'll proceed and create factories for the other models, so we'll end up with a clean interface for creating demo data for both seeds and tests._

```tsx
// It's simple like that
const randomPerson = personFactory.build();

// Also, we can build up multiple items
const randomPersons = personFactory.buildList(5);

// We can provide custom values for some of the fields
const adminPerson = personFactory.build({
  firstName: 'Admin',
  lastName: 'Dev',
  email: 'admin@podkrepi.bg',
})

// It can also handle relations between models..
```

https://github.com/thoughtbot/fishery


kachar (approved):
@valkirilov The used approach seems very intuitive to me and I support moving in this direction.

when we build our seeds with ease we can achieve various scenarios with less effort.

valkirilov:
**Progress update:** 
Created more factories to be used for data seeds:
- created factories for City, Country, CampaignType, Company, InfoRequest, Supporter, and Coordinator
- refactor the essential seeds to use factories and simplify seeding
- update seeds for the dev data to use factories (partially)

_Note: Thank you for your approval @kachar, I plan to create factories for the rest of the existing models (used in the seeds) as well, but I left the most complex to be the last ones. Feel free to merge this (unless you want to wait for more approvals), so we don't pile up huge PRs with a lot of changes, because it will become difficult to review. I can create new PRs when I'm ready with the changes for the other models._

igoychev:
Great improvement @valkirilov with the factory pattern!

> PS: Tests are failing here in the PR (maybe because of the forked repo?), but it's all green on my local:

Regarding the failing tests during github action - looking at the logs it seems you also need to commit the updated yarn.lock file after adding the fishery lib.

valkirilov:
> Regarding the failing tests during github action - looking at the logs it seems you also need to commit the updated yarn.lock file after adding the fishery lib.

Looking at the changelog it seems that the `yarn.lock` is updated properly and committed as part of this PR (or at least it contains a section with info related to fishery)

igoychev:
> > Regarding the failing tests during github action - looking at the logs it seems you also need to commit the updated yarn.lock file after adding the fishery lib.
> 
> Looking at the changelog it seems that the `yarn.lock` is updated properly and committed as part of this PR (or at least it contains a section with info related to fishery)

Well, it looks there is still an uncommitted change that is reported by the test run (see screenshot). May be you added fishery and committed the yarn.lock (which was ok), then removed the carret ^ from packages.json manually, but then didn't commit/push the yarn.lock again (which is where the diff comes from). 
Screenshot 2022-12-22 at 10 59 59

valkirilov:
Mmm good point. Maybe that was the case :thinking: 
Just updated again the `yarn.lock` file, but for some reason it's still failing :disappointed:

PR #404, opened 2022-12-18
filled billingName with 'cash payment' when OppositeSideName is empty

quantum-grit:
In case of cach payment there might be no name of the donor, so billingName will remain empty. 
For such cases better to fill it with 'вноска на каса', instead of leaving it empty.


igoychev (approved):

kachar (approved):
Nice update @quantum-grit - it's much better to show the type of transaction when we don't have opposite side.

PR #411, opened 2023-01-06
Added more factories for the seeds

valkirilov:
Introduce factories for the different seeds

- added factories for the Beneficiary, BankAccount, Vault, and Expense models
- minor refactoring to the seeds scripts

**Note:** These changes are a follow-up of the following PR: 
- https://github.com/podkrepi-bg/api/pull/403

_PS: There are two more entities that need to be updated to use the factories pattern, the most complex ones - Campaign and Donations, so I'll deal with them in a separate PR._


igoychev (approved):
beautiful code!

PR #415, opened 2023-01-10
Recurring donations credentials fix

slavcho:
There was a problem with the credentials.
The list of the recurring donations and the cancel operation were allowed only for admin users.

Also stopped the exceptions if an payment intent has no metadata. This currently happens with the subscriptions. Hence I am using the invoice.paid event.
This will probably be cleaned in a future fix.


igoychev (approved):
cool, all looks good

PR #421, opened 2023-01-23
Fixed double counting in vault.amount when same bank import is uploaded twice

quantum-grit:
This PR fixes a problem with incorrect increase of vault amount when importing bank donations from the same file twice.
The problem was with using upsert without checking if the record was already existing or not. Fixed by first trying to create the record and increase vault amounts only if the create is successful.

Full list of improvements:
- fixed seeds to not increment vault with not succesful donations
- fixed: SUM of returned currentAmount to not include blockedAmount so that Admin UI to display currentlyAvailableAmounts
- fixed bug of double counting when importing same bank donations
- corrected the existing vault amounts in database by updating with sum of donations


apps/api/src/campaign/campaign.service.ts
```diff
@@ -78,46 +78,36 @@ export class CampaignService {
...
+        GROUP BY source_vault_id
+      ) as w
+      ON w.source_vault_id = v.id
+    GROUP BY v.campaign_id
+    ${
+      campaignIds && campaignIds.length > 0
+        ? Prisma.sql`HAVING v.campaign_id::TEXT in (${Prisma.join(campaignIds)})`
+        : Prisma.empty
+    }`
+
```
slavcho, line 109: nice refactoring

apps/api/src/donations/donations.service.ts
```diff
@@ -423,17 +423,25 @@ export class DonationsService {
...
-          update: donation,
-          create: donation,
+        //to avoid incrementing vault amount twice we first try to create
+        await this.prisma.donation.create({
+          data: donation,
         })
 
-        //now update the Vault amounts too
+        //and if unique constraint on extPaymentIntentId fails this line will be skipped
         this.vaultService.incrementVaultAmount(donation.targetVaultId, donation.amount)
```
slavcho, line 432: I think relying on exceptions during the donation create is not best approach.
Here is my suggestion:

```
await prisma.$transaction(async (tx) => {
      const oldDonation = this.prisma.donation.findUnique(...);
      await this.prisma.donation.upsert({
          where: { extPaymentIntentId: donation.extPaymentIntentId },
          update: donation,
          create: donation,
      if (!oldDonation) {
            this.vaultService.incrementVaultAmount(donation.targetVaultId, donation.amount)
      } else {
                   where: { extPaymentIntentId: donation.extPaymentIntentId },
                    data: donation,
           })
   }
} );

```

slavcho (commented):
I think using a transaction would make the code more robust.

apps/api/src/donations/donations.service.ts
quantum-grit: Thanks for pointing me to a better  approach, done now! The code became better and today I learned how prisma interactive transactions work - love them!

slavcho (approved):

igoychev (approved):

PR #410, opened 2022-12-29
Payment intent endpoint

dimitur2204:
Add an endpoint for creating a PaymentIntent, which is needed to initialize the PaymentElement on the frontend. 
https://stripe.com/docs/api/payment_intents/create?lang=node

Related to Frontend https://github.com/podkrepi-bg/frontend/pull/1269

image
image

PR #425, opened 2023-01-29
407 - New Donation Flow PaymentIntent endpoints and flow 

dimitur2204:
Closes #409

## Motivation and context
This PR includes new endpoints needed for the new Donation Flow. Check out https://github.com/podkrepi-bg/frontend/issues/1272.

There are changes on how the Stripe donation flow works for the checkout session and the new Payment Intent approach.
This is the guide that was followed when creating these new endpoints: 
https://stripe.com/docs/payments/accept-a-payment?platform=web&ui=elements
## Testing

No testing done currently.
### Steps to test

Go to swagger at http://localhost:5010/swagger (they are public and you do not need a toke)
### New endpoints

- `@Post('payment-intent')` - Creates a new stripe payment intent based and returns it
- `@Post('payment-intent/:id')` - Updates a payment intent
- `@Post('create-stripe-payment')` - Creates a donation based on the intent info and data provided by the user through the form on the frontend

Note: The two `payment-intent` endpoints are both `POST`, because of how the Stripe API is setup. We can change that though (https://stripe.com/docs/api/payment_intents)


apps/api/src/donations/donations.service.ts
```diff
@@ -114,6 +115,87 @@ export class DonationsService {
...
 
+  /**
+   * Create initial donation object for tracking purposes
+   * This is used when the payment is created from the payment intent
+   */
+  async createInitialDonationFromIntent(
+    campaign: Campaign,
+    stripePaymentDto: CreateStripePaymentDto,
+    paymentIntent: Stripe.PaymentIntent,
+  ): Promise<Donation> {
```
dimitur2204, line 126: This is basically a copy of `createInitialDonationFromSession` adapted to work with the `stripePaymentDto` and the new requests.

apps/api/src/donations/donations.service.ts
```diff
@@ -114,6 +115,87 @@ export class DonationsService {
...
+    const targetVaultData = vault
+      ? // Connect the existing vault to this donation
+        { connect: { id: vault.id } }
+      : // Create new vault for the campaign
+        { create: { campaignId: campaign.id, currency: campaign.currency, name: campaign.title } }
+
+    /**
+     * Create or update initial donation object
+     */
+    const donation = await this.prisma.donation.upsert({
```
dimitur2204, line 146: This is an upsert call since I ran into `409 Conflict` errors, because when that method is called the `payment_intent.create` hook had already created a donation with that `extPaymentIntentId` and it should be unique for every donation.

Maybe it should be a straight `update` call since the frontend will call the `create-stripe-payment` always after the `payment_intent.created` has been sent

apps/api/src/donations/donations.controller.ts
```diff
@@ -152,7 +163,7 @@ export class DonationsController {
     return this.donationsService.create(createPaymentDto, user)
   }
 
-  @Post('create-payment-intent')
+  @Post('payment-intent')
```
kachar, line 166: Since we're deprecating the `/create-payment-intent` endpoint in favor of `/payment-intent` have we changed the current flow of the donations to use the new endpoint? 

I'm asking to be sure that we don't interrupt the donation process until v2 donation flow is set in place.

dimitur2204: The `create-payment-intent` endpoint is not used for the current donation flow. 
It was merged into the api master branch in a PR a week ago https://github.com/podkrepi-bg/api/pull/410

apps/api/src/donations/donations.service.ts
```diff
@@ -114,6 +115,87 @@ export class DonationsService {
...
+        amount: 0, //this will be updated on successful payment event
+        chargedAmount: paymentIntent.amount,
+        currency: campaign.currency,
+        provider: PaymentProvider.stripe,
+        type: DonationType.donation,
+        status: DonationStatus.initial,
+        extCustomerId: stripePaymentDto.personEmail ?? '',
+        extPaymentIntentId: paymentIntent.id,
+        extPaymentMethodId: 'card',
+        billingEmail: stripePaymentDto.isAnonymous ? stripePaymentDto.personEmail : null, //set the personal mail to billing which is not public field
```
kachar, line 158: Maybe we can add the billing email in all cases, not only whey user has selected anonymous donation.

apps/api/src/donations/donations.service.ts
```diff
@@ -114,6 +115,87 @@ export class DonationsService {
...
+    const donation = await this.prisma.donation.upsert({
+      where: { extPaymentIntentId: paymentIntent.id },
+      create: {
+        amount: 0, //this will be updated on successful payment event
+        chargedAmount: paymentIntent.amount,
+        currency: campaign.currency,
+        provider: PaymentProvider.stripe,
+        type: DonationType.donation,
+        status: DonationStatus.initial,
+        extCustomerId: stripePaymentDto.personEmail ?? '',
```
kachar, line 155: This looks like a data inconsistency - to store an email in the ID field.

Normally I think the values stored in this column should start with `cus_...`

dimitur2204: We should probably change it in the current implementation too then, because it is the same https://github.com/podkrepi-bg/api/pull/425/files/ba5b5f56e1d5c68b51284575b93fd9c0f06a9882#diff-3941d803ef3b8b94eac319c8965e8ea1a254e002f4f3878dae61f4b3ca783c15R82

apps/api/src/donations/donations.service.ts
dimitur2204: That would make sense, it would actually make more sense to not save the email if he has chosen `isAnonymous` if it is not required.
Then we might not be able to send the receipt though if this is the email that is used for that.

apps/api/src/donations/donations.service.ts
dimitur2204: I was thinking that this email is what connects the donation to the user

apps/api/src/donations/donations.service.ts
kachar: As you said the `billingEmail` is never exposed to the public, so it's always anonymous.

I'd suggest we keep the billing email so we can contact the person if there's a problem with the payments or other circumstances.

apps/api/src/donations/donations.service.ts
kachar: Looks like this field is a mess...

https://github.com/podkrepi-bg/api/blob/9d5e7a80c0d3ca2a4add449b9d7dc1b01fbb3b58/apps/api/src/bank-transactions-file/helpers/parser.ts#L27


https://github.com/podkrepi-bg/api/blob/54273881695cc1290ae9c3ba370f676ea8148024/apps/api/src/campaign/campaign.service.ts#L468

https://github.com/podkrepi-bg/api/blob/a89b7007edbc46c0151edad6c3095315a0077b27/apps/api/src/donations/events/stripe-payment.service.ts#L168

https://github.com/podkrepi-bg/api/blob/5b0589f3e162eee0034ef8a3acdf1b450339168b/apps/api/src/donations/donations.service.ts#L81

I'd say let's leave it as is and create a new issue to improve the situation and the data integrity.

dimitur2204: https://github.com/podkrepi-bg/api/issues/428

apps/api/src/donations/donations.service.ts
dimitur2204: [407 - always add billingEmail as the provided personEmail](https://github.com/podkrepi-bg/api/pull/425/commits/7db3a5a62923e26b4a3a0c38b48ede847cc62eb3)

kachar (approved):

PR #431, opened 2023-02-06
Feature/notify all users on donation

AndreyGoranov:
Closes [Issue](https://github.com/podkrepi-bg/frontend/issues/911)
Should go together with https://github.com/podkrepi-bg/frontend/pull/1325

---|---|---|---


AndreyGoranov:
> #### ⚠️ GitGuardian has uncovered 2 secrets following the scan of your pull request.
> Please consider investigating the findings and remediating the incidents. Failure to do so may lead to compromising the associated services or software components.
> 
> 🔎 Detected hardcoded secrets in your pull request
> 🛠 Guidelines to remediate hardcoded secrets
> 🦉 [GitGuardian](https://dashboard.gitguardian.com/auth/login/?utm_medium=checkruns&utm_source=github&utm_campaign=cr1) detects secrets in your source code to help developers and security teams secure the modern development process. You are seeing this because you or someone else with access to this repository has authorized GitGuardian to scan your pull request.Our GitHub checks need improvements? [Share your feedbacks](https://050127dayho.typeform.com/to/KmeAPTMk)!

The secret is revoked :)

apps/api/src/campaign/campaign.service.ts
```diff
@@ -408,10 +410,25 @@ export class CampaignService {
...
+      amount: true,
+      createdAt: true,
+      extPaymentMethodId: true,
+      person: {
+        select: {
+          firstName: true,
+          lastName: true,
+          picture: true,
+        },
+      },
```
kachar, line 426: @AndreyGoranov Is it expected to expose the data of the person who donated to all other website visitors?

We should restrict the personal data in the notifications based on the user's choice.

The user can select to stay anonymous when they donate, so we shouldn't expose that info to the general public.

cc @igoychev @dimitur2204

AndreyGoranov: How can I check if the user selected to stay Anonymous?

kachar: @AndreyGoranov I've just researched the topic and it seems that the `personId` of the user that selected `anonymous` is never sent to the backend, so I assume that the donations with a connected person are always non-anonymous.

Having that in mind I'll resolve the discussion.

igoychev: yes, when the user wants to donate anonymously, the donation record will not be connected to a person entity. So for anonymous donations, only the billingName of the donation will hold the name, and we don't display that publicly.

apps/api/src/campaign/campaign.service.ts
```diff
@@ -463,6 +481,8 @@ export class CampaignService {
             person: { connect: { id: paymentData.personId } },
           },
         })
+
+        this.webSocketService.sendNotification('successfulDonation', donation)
```
igoychev, line 485: here the donation is returned from a create statement and it will contain all donation fields. Could you please Pick<> only the needed fields

apps/api/src/campaign/campaign.service.ts
```diff
@@ -492,6 +512,11 @@ export class CampaignService {
             billingEmail: paymentData.billingEmail,
           },
         })
+
+        this.webSocketService.sendNotification('successfulDonation', {
+          ...updatedDonation,
+          person: donation.person,
+        })
```
igoychev, line 519: here the donation is returned from an Update statement and it will contain all donation fields. Could you please Pick<> only the needed fields for the notification

apps/api/src/sockets/socket.service.ts
```diff
@@ -0,0 +1,15 @@
+import { Injectable } from '@nestjs/common'
+import { SocketGateway } from './gateway'
+
+@Injectable()
+export class WebSocketService {
```
igoychev, line 5: [naming guideline] NotificationService would be a better name to show the purpose

apps/api/src/sockets/gateway.ts
```diff
@@ -0,0 +1,55 @@
...
+  WebSocketGateway,
+  WebSocketServer,
+} from '@nestjs/websockets'
+import { JwtService } from '@nestjs/jwt'
+import { PersonService } from '../person/person.service'
+import { Person } from '@prisma/client'
+import { Logger } from '@nestjs/common'
+
+@WebSocketGateway({ namespace: '/api/v1', transport: 'websocket' })
+export class SocketGateway implements OnGatewayConnection, OnGatewayInit, OnGatewayDisconnect {
```
igoychev, line 14: The general naming guideline is the name to show the intended usage of the class.
In this case:
The purpose of this gateway is to Notify, hence NotificationGateway is a better name. In contrast, a SocketGateway name shows only what it is made of, but not what it is used for.

apps/api/src/sockets/socket.module.ts
```diff
@@ -0,0 +1,20 @@
...
+    SocketGateway,
+    WebSocketService,
+    JwtService,
+    PersonService,
+    PrismaService,
+    ConfigService,
+  ],
+  exports: [WebSocketService],
+})
+export class WebSocketModule {}
```
igoychev, line 20: [naming guideline] NotificationModule would show the purpose of this module

apps/api/src/campaign/campaign.service.ts
igoychev, line 426: Could we declare the DonationNotificationSelect as a type in the NotificationService and then include it here from it?

igoychev (changes requested):
Great job overall! 
just please see comments to not notify with more data than needed and the naming guidelines.

I may have more comments, because haven't reviewed the socket gateway and connection logic

igoychev (approved):
great!

PR #433, opened 2023-02-07
Cancel Payment Intent Endpoint

dimitur2204:
## Motivation and context

Adds an enpoint for cancelling Payment Requests. Needed to do cleanup on cancelled donations in the new donation flow.

### New endpoints
`@Post('payment-intent/:id/cancel')` - 

#### Body:
```
cancellation_reason: Stripe.PaymentIntentCancelParams.CancellationReason
```


kachar (commented):
@dimitur2204 I see that the new endpoint is open to the public, so I wonder if we should allow canceling a payment intent of another person?

dimitur2204:
> @dimitur2204 I see that the new endpoint is open to the public, so I wonder if we should allow canceling a payment intent of another person?

This way if you are not logged in and you decide to cancel the payment you will not be able to. 
Generally speaking this cancellation is not required since Stripe just keeps them as `Incomplete` and probably gets rid of them after some time. And also we might not be able to reliably call it on the frontend because of the many ways a person can navigate away from the form

kachar (approved):

PR #436, opened 2023-02-10
New Donation Flow - Payment status bug fixes after testing

dimitur2204:
## Motivation and context
Added `metadata` param to the request for updating a payment intent to be able to attach it at the end of the donation flow

After doing some testing I saw that the status of the donation was not updated correctly. It turned out to be because of this check 
https://github.com/podkrepi-bg/api/pull/436/files#diff-88a7098916ffc30bd6ca323aa43535fbe9f8301dbc7a113af66c9b5bc8ec055bL478
https://github.com/podkrepi-bg/api/blob/a61b4997f6b93b88dc8cc0b75f2bd8c0cfe5465d/apps/api/src/donations/helpers/donation-status-updates.ts#L8
Which I can't exactly wrap my head around, it probably had something to do with how previously the stirpe session was updating the payments and the order in which it was doing it.

### DIFF
--- a/apps/api/src/campaign/campaign.service.ts
+++ b/apps/api/src/campaign/campaign.service.ts
@@ -23,7 +23,6 @@ import { VaultService } from '../vault/vault.service'
 import { CreateCampaignDto } from './dto/create-campaign.dto'
 import { UpdateCampaignDto } from './dto/update-campaign.dto'
 import { PaymentData } from '../donations/helpers/payment-intent-helpers'
-import { getAllowedPreviousStatus } from '../donations/helpers/donation-status-updates'
 import { Prisma } from '@prisma/client'
 import { CampaignSummaryDto } from './dto/campaign-summary.dto'
 import {
--- a/apps/api/src/campaign/campaign.service.ts
+++ b/apps/api/src/campaign/campaign.service.ts
@@ -1,4 +1,5 @@
 import {
+  Prisma,
   Campaign,
   CampaignState,
   CampaignType,
@@ -20,10 +21,10 @@ import {
 import { PersonService } from '../person/person.service'
 import { PrismaService } from '../prisma/prisma.service'
 import { VaultService } from '../vault/vault.service'
+import { shouldAllowStatusChange } from '../donations/helpers/donation-status-updates'
+import { PaymentData } from '../donations/helpers/payment-intent-helpers'
 import { CreateCampaignDto } from './dto/create-campaign.dto'
 import { UpdateCampaignDto } from './dto/update-campaign.dto'
-import { PaymentData } from '../donations/helpers/payment-intent-helpers'
-import { Prisma } from '@prisma/client'
 import { CampaignSummaryDto } from './dto/campaign-summary.dto'
 import {
   AdminCampaignListItem,
@@ -472,7 +473,10 @@ export class CampaignService {
       return
     }
     //donation exists, so check if it is safe to update it
-    else if (donation?.status !== newDonationStatus) {
+    else if (
+      donation?.status !== newDonationStatus &&
+      shouldAllowStatusChange(donation?.status, newDonationStatus)
+    ) {
       try {
         await this.prisma.donation.update({
           where: {
--- a/apps/api/src/campaign/campaign.service.ts
+++ b/apps/api/src/campaign/campaign.service.ts
@@ -473,10 +473,7 @@ export class CampaignService {
       return
     }
     //donation exists, so check if it is safe to update it
-    else if (
-      donation?.status !== newDonationStatus &&
-      shouldAllowStatusChange(donation?.status, newDonationStatus)
-    ) {
+    else if (shouldAllowStatusChange(donation.status, newDonationStatus)) {
       try {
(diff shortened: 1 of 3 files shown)


kachar (commented):
I'm not really sure why the previous state check exists so I can't say if it's okay to remove it.

Maybe @igoychev can recall info on this decision

sashko9807:
> I'm not really sure why the previous state check exists so I can't say if it's okay to remove it.

[#300](https://github.com/podkrepi-bg/api/pull/300)

apps/api/src/donations/helpers/donation-status-updates.ts
```diff
@@ -1,21 +1,50 @@
...
+): boolean {
+  if (isFinal(oldStatus)) {
+    return false
+  }
+
+  if (
+    (isFinal(newStatus) || isChangeable(newStatus)) &&
+    (isChangeable(oldStatus) || isInitial(oldStatus))
+  ) {
+    return true
```
dimitur2204, line 46: @kachar @igoychev Does that make more sense like that?

apps/api/src/donations/helpers/donation-status-updates.ts
```diff
@@ -1,21 +1,50 @@
...
   }
-  return undefined
+
+  throw new Error(`Unhandled donation status change from ${oldStatus} to ${newStatus}`)
 }
```
dimitur2204, line 50: @kachar @igoychev Does that make more sense like that?

apps/api/src/campaign/campaign.service.ts
```diff
@@ -474,8 +474,8 @@ export class CampaignService {
     }
     //donation exists, so check if it is safe to update it
     else if (
-      donation?.status === newDonationStatus ||
-      donation?.status === getAllowedPreviousStatus(newDonationStatus)
+      donation?.status !== newDonationStatus &&
+      shouldAllowStatusChange(donation?.status, newDonationStatus)
```
igoychev, line 478: The previous implementation allowed the donation to be updated if the old and new status was the same. 
That was to allow replay of old events if we want to add more fields from the event.

apps/api/src/donations/helpers/donation-status-updates.ts
```diff
@@ -1,21 +1,50 @@
 import { DonationStatus } from '@prisma/client'
 
+const initial: DonationStatus[] = [DonationStatus.initial]
+const changeable: DonationStatus[] = [
+  DonationStatus.incomplete,
+  DonationStatus.paymentRequested,
+  DonationStatus.waiting,
```
igoychev, line 7: it looks to me, initial should also be changeable

apps/api/src/donations/helpers/donation-status-updates.ts
```diff
@@ -1,21 +1,50 @@
...
-    case DonationStatus.cancelled: {
-      return DonationStatus.waiting
-    }
+export function shouldAllowStatusChange(
+  oldStatus: DonationStatus,
+  newStatus: DonationStatus,
+): boolean {
+  if (isFinal(oldStatus) || isInitial(newStatus)) {
+    return false
+  }
```
igoychev, line 44: updating when old === new should also be possible

igoychev (changes requested):
I like the change, however please allow same status to be update-able as it was with the previous logic

apps/api/src/donations/helpers/donation-status-updates.ts
dimitur2204: Check [add a check on shouldAllowStatusChange to return true when the st…](https://github.com/podkrepi-bg/api/pull/436/commits/73ad3b4c3295c3932053380f1d58d37d5038ca28)

apps/api/src/donations/helpers/donation-status-updates.ts
```diff
@@ -35,6 +35,10 @@ export function shouldAllowStatusChange(
   oldStatus: DonationStatus,
   newStatus: DonationStatus,
 ): boolean {
+  if (oldStatus === newStatus) {
```
sashko9807, line 38: I might be be missing something, but wouldn't short circuit make this check redundant ? What I mean is that whenever 
[donation?.status !== newDonationStatus ](https://github.com/podkrepi-bg/api/pull/436/commits/1f7c1c2dcab124771081edf9a9e023d7e12763a1#diff-88a7098916ffc30bd6ca323aa43535fbe9f8301dbc7a113af66c9b5bc8ec055bR477) evaluates to false, due to usage of  AND operator, shouldAllowStatusChange() shouldn't be called at all.

dimitur2204: That is true actually. Nice catch! That would not get called at all in that case. 
I think I will leave it though in the case this function is used somewhere else in the future. I think this check makes sense.

igoychev: @dimitur2204 yep this could be left here, however please check to remove the donation?.status !== newDonationStatus where the function is called

igoychev (approved):

PR #437, opened 2023-02-13
Modified Prisma UniqueConstraint Error Structure returned in response

BogoCvetkov:
This PR is connected to this PR on the FE [#1341](https://github.com/podkrepi-bg/frontend/pull/1341)

## Motivation and context

Handling of unique constraint violation errors was implemented on the FE, so that the users receive informative messages why the submission failed. Currently the FE expects the structure of the validation errors (that are thrown if the payload is not in the correct format) to display error messages. I re-modelled the structure of the 409 errors that Prisma throws on Unique Constraint Violation in the same way as validation errors so that the logic for extracting the error messages on the FE remains the same.

This change is required so that, the implemented error handling for unique constraint violation on the FE works


igoychev (approved):
nice improvement - great to see you found the right place to make it!

PR #439, opened 2023-02-14
related to new avatars

martin23535:
Api changes related to new avatars in the front end


igoychev (approved):

apps/api/src/auth/dto/register.dto.ts
```diff
@@ -26,4 +26,10 @@ export class RegisterDto {
   @IsNotEmpty()
   @IsString()
   public readonly lastName: string
+
+  @ApiProperty()
+  @Expose()
+  @IsNotEmpty()
+  @IsString()
+  public readonly picture: string
```
igoychev, line 34: I think the picture should be optional

igoychev:
this is stalled for long time

PR #440, opened 2023-02-19
Enhancement/make older slugs redirect to campaign after update

BogoCvetkov:
### Why is this change required? 
Currently when a campaign slug is updated with a new one, the old one just stops existing. As a consequence all the campaign urls that contain the old slug stop working. This means that all the links that have been shared previously (friends, social media, etc. ) are invalidated. It would be much better if the platform keeps track of old slugs, so that it's easier for people to find the campaign, long after the link to it has changed

### How did you solve the problem? 
1. I have created a new table in the DB called `SlugArchive`, which keeps track of all slugs that were created in the platform and that are not currently used in any active campaign(meaning they were replaced by a new slug).
   - Each record in the table holds the `slug name` and the `campaign id` it used to point to, before it was replaced(updated)
2. Updated the `getCampaignBySlug` method inside `campaign.service` so that is uses the new `Slug Archive`
   - When the route for this method is called, first a check is made in the `Campaigns Table` to check if any campaign currently uses it (as was the original behaviour). 
   - Now instead of throwing a `Not found exception` directly after that, it conducts one additional check in the `SlugArchive Table` to see it this slug was not used in the past for some campaign.
 3. Updated the `update` method inside `campaign.service` so that is uses the new `Slug Archive`
    - After the changes are applied to the DB and the record is updated an additional step is added - a check is made if the `updateDto` includes the slug and if the slug is a new one. If the check passes, a new record is added to the `SlugArchive table`. In case a record for this slug already  exists it just gets updated to point to the campaign that was last using it.
 4. Updated the tests to reflect the new additions
 
### How it should work in practice ?
1. Let's say we create a campaign with a slug `initial-slug` and a name `Campaign 1`
2. We update  `Campaign 1`  slug to `updated-slug`
3. Now both  `initial-slug` &  `updated-slug` point to `Campaign 1`
4. Now we create a second campaign with a slug `initial-slug` and a name `Campaign 2` 
    - NOTE:  `initial-slug` now exists only in the `SlugArchive Table` which means that it's free to be used as a current slug by any campaign. This means we could assign it to `Campaign 2` without violating the unique constraint
5. Now  `updated-slug` points to `Campaign 1` &   `initial-slug` points to `Campaign 2` (because we always check for the current slugs first)
6. We update  `Campaign 2`  slug to `updated-slug-2`
7. Now `updated-slug` => `Campaign 1` ,   `updated-slug-2` =>  `Campaign 2`,  `initial-slug` =>  `Campaign 2` (there was already a record in the `SlugArchive Table`, we just updated it to the campaign that used it last)

### NOTE:
 - Slugs are tracked only for campaigns in `active state` (which were in the `active state` before and after the update).  Since those are the ones visible to the users and we don't want to store unnecessary data in the DB.


schema.prisma
```diff
@@ -209,10 +209,22 @@ model Campaign {
...
   withdrawals             Withdrawal[]
+  slugArchive             SlugArchive[]
 
   @@map("campaigns")
 }
 
+/// Keeps track of previous slugs that are not used currently in any active campaign
+model SlugArchive {
+ id                      String         @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
+ slug                    String         @unique @db.VarChar(250)
```
igoychev, line 220: if the slug is UNIQUE then let's make it PRIMARY and we can go without the "id" column

igoychev (changes requested):
looks great and well done with adding a test for this!

I am just requesting a minor change for the seemingly unnecessary id field in the new archive table

schema.prisma
BogoCvetkov: I have applied those changes

igoychev (approved):

PR #445, opened 2023-02-20
Add Startup Probe to allow container to start

quantum-grit:
## Motivation and context
Recently the deployments are failing on readiness healthcheck with 503 service not available.
To give more time to start, this PR adds startupProbe with up to 2min (12 attempts per 10sec) before enabling the readiness and startup probes.


imilchev (approved):
LGTM

PR #447, opened 2023-02-21
[Bug] Payment import: User feedback message on donation import operations

Nnachevvv:
The requirement for this PR is that when we import bank transaction we don't provide any feedback to the admin.
Currently we have two messages for feedback after payments import - "ok" and "error". This information is not enough.
I did a refactoring to the bank-transaction-file.controller.ts to returns a status of the imported records to the frontend.

Relates to https://github.com/podkrepi-bg/frontend/pull/1366 and closes https://github.com/podkrepi-bg/frontend/issues/1137


apps/api/src/donations/donations.service.ts
```diff
@@ -546,29 +547,33 @@ export class DonationsService {
     return donation
   }
 
-  async createManyBankPayments(donationsDto: CreateManyBankPaymentsDto[]) {
-    for (const donation of donationsDto) {
-      await this.prisma.$transaction(async (tx) => {
-        //to avoid incrementing vault amount twice we first check if there is such donation
-        const existingDonation = await tx.donation.findUnique({
-          where: { extPaymentIntentId: donation.extPaymentIntentId },
+  async createUpdateBankPayment(donationsDto: CreateManyBankPaymentsDto): Promise<ImportStatus> {
```
igoychev, line 550: I see the function is refactored to work on a single donation - could you rename the input parameter from plural to singular too

apps/api/src/bank-transactions-file/bank-transactions-file.controller.ts
```diff
@@ -75,32 +76,47 @@ export class BankTransactionsFileController {
...
-        movement.payment.extPaymentMethodId = 'imported bank payment'
-        movement.payment.targetVaultId = vault[0].id
-        movement.payment.type = DonationType.donation
-        movement.payment.status = DonationStatus.succeeded
-        movement.payment.provider = PaymentProvider.bank
-        donations.push(movement.payment)
+      try {
+        donationImportRecord.status = await this.donationsService.createUpdateBankPayment(
+          movement.payment,
+        )
```
igoychev, line 110: well done with the refactoring of processing the donations in the one cycle  - the previous code was more complex and creating additional collection that was cycled separately

igoychev (changes requested):
Many people will buy you many beers for this update! It saves so much manual efforts!

All good with the changes, just please address the naming and logging comments below.

apps/api/src/donations/donations.service.ts
Nnachevvv: I agree that this should be renamed. We have a similar object **CreateBankPaymentDto** https://github.com/podkrepi-bg/api/blob/master/apps/api/src/donations/dto/create-bank-payment.dto.ts . 
From my investigation , I saw that we expose it through the following endpoint `/create-bank-payment`, but I couldn't find where it is used. My assumption is that the object is outdated. 
If this is the case I can refactor **CreateBankPaymentDto** with the proper fields and reuse it instead **CreateManyBankPaymentsDto**. Would you please confirm if this object and endpoint is not in use?

apps/api/src/bank-transactions-file/dto/bank-transactions-import-status.dto.ts
```diff
@@ -0,0 +1,17 @@
+import { DateTime } from 'aws-sdk/clients/devicefarm'
```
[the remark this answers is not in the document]
Nnachevvv: This was strange to me too, I saw that we are using this library in DTO definition  apps/api/src/donations/dto/create-many-bank-payments.dto.ts, therefore I applied that here. I changed DateTime to Date in both places

apps/api/src/donations/donations.service.ts
igoychev: well spotted! 
The CreateBankPaymentDto and  CreateManyBankPaymentsDto should be the same as they are used for the same thing. Indeed the CreateBankPaymentDto is outdated - it is used for manual creation of bankImports in the same Admin UI where the file import happens. Please take the new fields from CreateManyBankPaymentsDto and let's unify them into one.
Btw. feel free to do this in separate PR because it will probably require more changes in frontend too

apps/api/src/bank-transactions-file/dto/bank-transactions-import-status.dto.ts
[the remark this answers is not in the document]
igoychev: great that you fixed the old imports too

igoychev (approved):

apps/api/src/donations/donations.service.ts
Nnachevvv: Alright! I would make a new PR after this one is  being merged, where I will unify the structures.

PR #449, opened 2023-02-22
Add Postgres adapter for socket communication

AndreyGoranov:
Since we have more than one server, we need this adapter as a way to notify all the servers of a specific event, so all the clients get notified. 
For example, we receive a valid payment on only one server, and we need a way to notify all the users who might be connected to the other server.


imilchev:
Please provide context about why this is needed and what purpose does it serve

imilchev (changes requested):
Would like to know more about this before it gets merged

AndreyGoranov:
> Would like to know more about this before it gets merged

https://docs.nestjs.com/websockets/adapter
https://socket.io/docs/v4/postgres-adapter/

apps/api/src/sockets/notifications/adapter.ts
```diff
@@ -0,0 +1,45 @@
...
+    await pool.connect()
+
+    await pool
+      .query(
+        `
+  CREATE TABLE IF NOT EXISTS socket_io_attachments (
+      id          bigserial UNIQUE,
+      created_at  timestamptz DEFAULT NOW(),
+      payload     bytea
+  );
```
kachar, line 30: The tables normally should be created via `schema.prisma` instead of at runtime

AndreyGoranov: Good to know. I will change it, thanks!

sashko9807:
If i correctly understood the [socket.io adapter docs](https://socket.io/docs/v4/adapter/), this is only needed in cases when we want to emit messages across multiple server instances, which does not seem to be the case for us, at least for now.

PR #452, opened 2023-03-04
Refactor environment variable configuration

### LINKED ISSUE #267: Remove .env files from source control
.env files are meant to be used on the individual local env and keeping them out of source control is a good practice. Makes the local setup much easier due to the fact that some services and debuggers rely on the .env file and ignore any overloads.

For the deployment environments, there is always a better way to provide the env vars (yml manifests, ci/cd variables, etc..)

valkirilov:
Based on the discussion under #267 I propose a variant with some minor refactoring 

- trying to follow the best practices, this commit removes the `.env` file from the repository to avoid exposing sensitive information. Instead, it adds a `.env.example` file that serves as a template for the required environment variables and their default values
- the README has also been updated to reflect the new changes and provide clear instructions for setting up the environment variables

Closes #267

### Note for DevOps:
I would recommend reviewing this pull request carefully as it has the potential to impact future deployments if the pipeline relies on the presence of a `.env` file in the repository. While I believe that environment variables should be configured based on their respective environments, it is important to take note of this potential issue and proceed with caution.

## Motivation and context

It is not a good idea to have the environment variables configuration file part of the repository via source control for several reasons:

- **Security**: Environment variables often contain sensitive information such as API keys, passwords, and other credentials. By including them in the source control, you are exposing this information to anyone who has access to the repository, increasing the risk of unauthorized access to sensitive data.

- **Different Configurations**: In many cases, developers may need to use different configurations based on their local environment, which may include different API keys, database URLs, or other environment-specific settings. By including a global configuration file in the repository, it makes it harder for developers to customize their environment.

- **Maintenance**: Keeping environment variables in the source control can lead to maintenance issues as new environment variables are added or old ones are removed. This could result in conflicts, merge issues, or other problems during code reviews or deployments.

Therefore, it's better to have environment variables configured outside of the source control and use configuration files like `.env.example` to provide guidance on default values for environment variables. Developers can create their own local configuration files like `.env.local` to override the global configurations while keeping their local changes separate from the default values.


kachar (commented):
We need to alter the running workflows in `.github/` folder in order to apply the new copy step to the action.

At the moment the API / DB tests are failing

valkirilov:
> We need to alter the running workflows in `.github/` folder in order to apply the new copy step to the action.
> 
> At the moment the API / DB tests are failing
> 
> ```yaml
> - name: Copy env file
>   run: cp .env.example .env
> ```
> 
> https://github.com/podkrepi-bg/api/blob/9348346d731052f9f5bb7845b2d8ea82f7c72029/.github/workflows/tests.yml#L30-L31

Good point, thank you! I managed to update the flow and the tests passed successfully, but I don't feel that confident when touching the GitHub Actions scripts, so still I would like to wait for @imilchev to double check 😄

kachar (approved):

PR #453, opened 2023-03-12
Backend support for the campaigns' expenses

slavcho:
We allow the user to add, fetch and delete campaign expenses.
Only the organizer (or admin) should be able to edit them. 
Only an admin can approve an expense.
But the general public should be able to see any approved expenses.
The expense files are uploaded to AWS.

## Environment

We need to create one new AWS bucket: 'expenses-files'


apps/api/src/expenses/expenses.controller.ts
```diff
@@ -22,12 +34,16 @@ export class ExpensesController {
   }
 
   @Post('create-expense')
+  @UseInterceptors(FilesInterceptor('file', 5, { limits: { fileSize: 10485760 } })) //limit uploaded files to 5 at once and 10MB each
```
dimitur2204, line 37: Do both the `/create-expense` and `/upload-files` endpoints need the functionality of accepting files. Are they called independently?

apps/api/src/campaign/campaign.controller.ts
```diff
@@ -55,6 +55,18 @@ export class CampaignController {
...
+  @Get('can-edit/:slug')
+  async canEditCampaign(
+    @Param('slug') slug: string,
+    @AuthenticatedUser() user: KeycloakTokenParsed,
+  ) {
+    const isOwner = await this.campaignService.isUserCampaign(user.sub, slug)
+
+    return isOwner || isAdmin(user)
+  }
+
```
dimitur2204, line 69: I don't think we need an endpoint for that, a [Guard](https://docs.nestjs.com/guards) on the endpoints that need editing permissions will probably do the job.

dimitur2204 (changes requested):
Amazing job 🚀! Comments are regarding security and authentication mostly

apps/api/src/expenses/expenses.controller.spec.ts
```diff
@@ -97,7 +102,7 @@ describe('ExpensesController', () => {
 
       const createDto: CreateExpenseDto = { ...expense }
 
-      await expect(controller.create(createDto)).rejects.toThrow()
+      // await expect(controller.create(createDto, [])).rejects.toThrow()
```
igoychev, line 105: if this is intentionally commented out - please add why it should stay so

apps/api/src/expenses/expenses.controller.ts
```diff
@@ -46,11 +62,60 @@ export class ExpensesController {
...
-    return this.expensesService.remove(id)
+  async listCampaignExpenses(@Param('slug') slug: string) {
+    return this.expensesService.listCampaignExpenses(slug)
+  }
+
+  @Get('campaign/approved/:slug')
+  @Public()
+  async listCampaignApprovedExpenses(@Param('slug') slug: string) {
+    return this.expensesService.listCampaignApprovedExpenses(slug)
+  }
```
igoychev, line 82: the route name could be more rest-like if named:
expenses/campaign/:slug/approved to indicate that the slug is of campaign, because currently it reads as if only approved campaigns will be returned.

apps/api/src/expenses/expenses.controller.ts
```diff
@@ -46,11 +62,60 @@ export class ExpensesController {
...
+
+  @Post('upload-files/:id')
```
igoychev, line 84: the upload-files/:id looks strange and it is not clear if this is a fileID or expenseID.
If expenseId then the route should be expenses/:id/files and the POST will instruct it is uploading

apps/api/src/expenses/expenses.controller.ts
```diff
@@ -46,11 +62,60 @@ export class ExpensesController {
...
+  @Get('campaign/:slug')
   @Roles({
     roles: [RealmViewSupporters.role, ViewSupporters.role],
     mode: RoleMatchingMode.ANY,
   })
-  remove(@Param('id') id: string) {
-    return this.expensesService.remove(id)
+  async listCampaignExpenses(@Param('slug') slug: string) {
+    return this.expensesService.listCampaignExpenses(slug)
+  }
```
igoychev, line 76: in general I would move the listCampaignExpenses to the campaign controler - so that the full route is campaigns/:slug/expenses

apps/api/src/expenses/expenses.controller.ts
```diff
@@ -46,11 +62,60 @@ export class ExpensesController {
...
+    mode: RoleMatchingMode.ANY,
+  })
+  uploadFiles(@Param('id') id: string, @UploadedFiles() files: Express.Multer.File[]) {
+    return this.expensesService.uploadFiles(id, files)
+  }
+
+  @Get('files/:id')
+  getUploadedFiles(@Param('id') id: string) {
+    return this.expensesService.listUploadedFiles(id)
+  }
```
igoychev, line 97: endpoint should be :id/files

apps/api/src/expenses/expenses.controller.ts
```diff
@@ -46,11 +62,60 @@ export class ExpensesController {
...
+
+  @Get('download-file/:fileId')
+  @Public()
+  async downloadFile(
+    @Param('fileId') fileId: string,
+    @Response({ passthrough: true }) res,
+  ): Promise<StreamableFile> {
```
igoychev, line 104: endpoint should be file/:fileId

apps/api/src/expenses/expenses.controller.ts
```diff
@@ -46,11 +62,60 @@ export class ExpensesController {
...
+    const file = await this.expensesService.downloadFile(fileId)
+    res.set({
+      'Content-Type': file.mimetype,
+      'Content-Disposition': 'attachment; filename="' + file.filename + '"',
+    })
+    return new StreamableFile(file.stream)
+  }
+
+  @Delete('file/:id')
```
igoychev, line 113: for consistency with the Get Above the endpoint should be: file/:fileId

igoychev (changes requested):

apps/api/src/campaign/campaign.controller.ts
igoychev: I think he needs it for knowing whether to show the Expense edit button. 
The call to backend could be avoided using the isAdmin function on the frontend which checks the roles of the logged user.

Btw. currently this shows every few seconds if the user is not logged in - so please fix before we merge it:
WARN [HTTP] GET /api/v1/campaign/can-edit/facilis-eveniet-maioresheavily-funded 401 Unauthorized

apps/api/src/expenses/expenses.service.ts
```diff
@@ -122,4 +130,66 @@ export class ExpensesService {
...
+    filename: string,
+    // uploadedBy: string,
+    buffer: Buffer,
+  ): Promise<string> {
+    const file: CreateExpenseFileDto = {
+      filename,
+      mimetype,
+      expenseId,
+      uploaderId: '9b26b247-c275-4320-9831-fb7a0d243758' /*TODO: get user id from context*/,
+    }
```
igoychev, line 146: this fails when testing locally - the easiest is to get it from @AuthenticatedUser() annotation - see campaign-file.controller.ts line 43

apps/api/src/expenses/expenses.service.ts
```diff
@@ -122,4 +130,66 @@ export class ExpensesService {
     const [result] = await this.prisma.$transaction([writeExpense, writeVault])
     return result
   }
+
+  async create_expense_file(
```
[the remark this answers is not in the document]
slavcho: fixed

apps/api/src/expenses/expenses.controller.ts
slavcho: You are right. I cleaned it.

apps/api/src/campaign/campaign.controller.ts
slavcho: I removed the can-edit endpoint. 
You were right, we can check if the current user is an organizer or admin in the frontend itself with the already available endpoints.

apps/api/src/expenses/expenses.controller.spec.ts
slavcho: OK. We don't validate the balance at the moment.
May be in the future if we need it - we can add it.

apps/api/src/expenses/expenses.controller.ts
slavcho: well the actual query from the front-end is like this:

/expenses/campaign/approved/:slug

The prefix is used to detect the controller.
I will leave it like this.

apps/api/src/expenses/expenses.controller.ts
slavcho: ok. fixed it.

apps/api/src/expenses/expenses.controller.ts
slavcho: fixed it.

apps/api/src/expenses/expenses.controller.ts
slavcho: I think download-file is better here. It is more verbose and explains that we are fetching the actual file's content, not the expense_file db record.

apps/api/src/expenses/expenses.controller.ts
slavcho: fixed it

apps/api/src/expenses/expenses.service.ts
slavcho: Sorry about this. fixed it already, but it was not pushed.

apps/api/src/expenses/expenses.controller.ts
slavcho: moved it.

apps/api/src/expenses/expenses.controller.ts
```diff
@@ -46,11 +62,60 @@ export class ExpensesController {
...
+  async listCampaignApprovedExpenses(@Param('slug') slug: string) {
+    return this.expensesService.listCampaignApprovedExpenses(slug)
+  }
+
+  @Post('upload-files/:id')
+  @UseInterceptors(FilesInterceptor('file', 5, { limits: { fileSize: 10485760 } })) //limit uploaded files to 5 at once and 10MB each
+  @Roles({
+    roles: [RealmViewSupporters.role, ViewSupporters.role],
+    mode: RoleMatchingMode.ANY,
+  })
```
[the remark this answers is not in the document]
slavcho: Actually - I need to check the owner of campaign (or admin). So a new role will not help.
We don't want one organizer to be able to change the expenses of a campaign than is not his (or hers).

I will do the changes.

apps/api/src/expenses/expenses.service.ts
```diff
@@ -122,4 +130,66 @@ export class ExpensesService {
...
+    expenseId: string,
+    mimetype: string,
+    filename: string,
+    // uploadedBy: string,
+    buffer: Buffer,
```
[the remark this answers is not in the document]
slavcho: fixed.

igoychev (approved):

Issue #466, opened 2023-03-16
[Refactor] Clean up donation creation flow


dimitur2204:
### Describe the feature you'd like to request

After recently having problems with the Stripe integration it seems like we have a lot of places to improve on the flow of data when creating a donation.

Related issues:
https://github.com/podkrepi-bg/api/issues/428
https://github.com/podkrepi-bg/api/issues/463

### Describe the solution you'd like

Think about a better more reusable and understandable approach on handling donations. This can come together with https://github.com/podkrepi-bg/frontend/pull/1324, so that the whole flow both frontend and backend gets refactored.

PR #467, opened 2023-03-16
#464 - Add check for an existing `person` and connect before updating `donation`

### LINKED ISSUE #464: [Bug] All donations are showing as anonymous

_No response_

### Describe the Bug

Even if you donate while logged in the donation still registers as anonymous.

### To Reproduce

1. Go to https://dev.podkrepi.bg/
2. Login to an account
3. Make a donation with the Stripe test data
4. Go back to the campaign page and check the donations card

### Expected Behavior

When you are logged in donations should be showing up as done by you in the donations card

### Which browser are you using? (if relevant)

_No response_

dimitur2204:
Closes #464

## Motivation and context

Quick fix to check if a person exists and then connect to the donation accordingly.
This whole flow needs revisiting and will be addressed in #466 

## Testing

Tested out the whole flow for both logged, anonymous and not logged and it is working as expected

### DIFF
--- a/apps/api/src/campaign/campaign.service.ts
+++ b/apps/api/src/campaign/campaign.service.ts
@@ -460,7 +460,12 @@ export class CampaignService {
 
       Logger.debug('Donation found by subscription: ', donation)
     }
-
+    let person
+    if (paymentData.billingEmail) {
+      person = await this.prisma.person.findFirst({
+        where: { email: paymentData.billingEmail },
+      })
+    }
     //if missing create the donation with the incoming status
     if (!donation) {
       Logger.debug(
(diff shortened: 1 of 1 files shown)


apps/api/src/campaign/campaign.service.ts
```diff
@@ -529,38 +541,13 @@ export class CampaignService {
...
-          },
-        })
-      }
-
-      Logger.debug('Saving donation wish ' + metadata?.wish)
-
-      if (metadata?.wish) {
-        await this.createDonationWish(metadata.wish, donation.id, campaign.id)
-      }
-    }
```
igoychev, line 563: The removed code was intended to be executed for Charge.Succeeded event so that only then to save the donation wish. Unfortunately moving it above doesn't guarantee that. So, please see my comment above that the actual bug is because this code is skipped by the forgotten 'return' on line 500.

igoychev (changes requested):

PR #483, opened 2023-04-03
Modify import bank transactions task filter

BogoCvetkov:
I've modified the background task service logic for importing bank transactions from IRIS. I've changed the filter so that only the bank donations get imported, instead of all transactions in the bank account (Outgoing, Stripe etc.)

- Some filtrations should be synced with the team, before this PR is ready for review


igoychev (approved):
well done with the filters!

PR #485, opened 2023-04-05
add iris config to kustomize

imilchev:
Added the necessary config for the IRIS integration. Note that this is not applied and will require a manual action first before it starts working

### DIFF
--- a/manifests/base/deployment.yaml
+++ b/manifests/base/deployment.yaml
@@ -113,6 +113,24 @@ spec:
                   name: jwt-secret
             - name: S3_BUCKET
               value: campaign-files
+            - name: IRIS_API_URL
+              value: https://developer.sandbox.irispay.bg/api/8
+            - name: BANK_BIC
+              value: UNCRBGSF
+            - name: PLATFORM_IBAN
+              value: test
+            - name: IMPORT_TRX_TASK_INTERVAL_MINUTES
+              value: '60'
+            - name: IRIS_AGENT_HASH
+              valueFrom:
+                secretKeyRef:
+                  name: iris-secret
+                  key: agent-hash
+            - name: IRIS_USER_HASH
+              valueFrom:
+                secretKeyRef:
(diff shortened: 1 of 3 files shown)


manifests/base/deployment.yaml
```diff
@@ -113,6 +113,24 @@ spec:
                   name: jwt-secret
             - name: S3_BUCKET
               value: campaign-files
+            - name: IRIS_API_URL
+              value: https://developer.sandbox.irispay.bg/api/8
```
imilchev, line 117: Since this is for dev, I assume there is another URL for the real integration

manifests/base/deployment.yaml
```diff
@@ -113,6 +113,24 @@ spec:
...
+            - name: BANK_BIC
+              value: UNCRBGSF
+            - name: PLATFORM_IBAN
+              value: test
```
imilchev, line 121: can we already add the correct IBAN? I assume that's the IBAN of the organization, which should be public anyways

manifests/base/deployment.yaml
```diff
@@ -113,6 +113,24 @@ spec:
...
+            - name: IMPORT_TRX_TASK_INTERVAL_MINUTES
+              value: '60'
+            - name: IRIS_AGENT_HASH
+              valueFrom:
+                secretKeyRef:
+                  name: iris-secret
+                  key: agent-hash
+            - name: IRIS_USER_HASH
```
imilchev, line 129: Can you confirm that both agent and user hash are supposed to be secrets?

manifests/base/deployment.yaml
igoychev: yes, BG66UNCR70001524349032

manifests/base/deployment.yaml
igoychev: yes both are private secrets

PR #487, opened 2023-04-18
Add platform statistics

Nnachevvv:
Part of https://github.com/podkrepi-bg/frontend/pull/1412, which will close this issue https://github.com/podkrepi-bg/frontend/issues/1391

## Motivation and context
In order to gather the statistics for the new Homepage section we would need a way to get the sum of donated money and number of the donors.

### Steps to test
GET request to:
http://localhost:3040/api/v1/donation/users-donated
http://localhost:3040/api/v1/vault/money
### New endpoints

This change opens two new endpoints needed for homepage statistics:
- vault/money - gather money from all vaults
- donations/users-donated - get number of donated users + anonymous users


apps/api/src/donations/donations.controller.ts
```diff
@@ -94,6 +94,12 @@ export class DonationsController {
     return await this.donationsService.getDonationsByUser(user.sub)
   }
 
+  @Get('users-donated')
+  @Public()
+  async usersDonated() {
```
igoychev, line 99: [naming] in this way the endpoint will be /api/v1/donation/user-donations, however it actually returs user-count, so let's name it like this: /api/v1/donation/donors-count

apps/api/src/donations/donations.service.ts
```diff
@@ -642,6 +642,20 @@ export class DonationsService {
     return user.id
   }
 
+  async getDonatedUsersCount() {
+    const donatedUsersCount = await this.prisma.donation.groupBy({
```
sashko9807: Can't quite get the idea to group the donations by personId. Is there a reason not to use the prisma count() API, to get the total amount of succeeded donations?

Nnachevvv: Yeah, After a discussion with @igoychev, we decided that it would be better to group the donations by "billingName". 

Regarding your question - we should have a way to count how many donations  has "billingName" as null value, since we count them as different donors. Unfortunately, this is currently not possible with count()+distinct functionality provided by prisma, at least I couldn't find out how to do that. Therefore we had a two ways to implement this - having two queries or the way which I had implemented it. I'm open to any suggestions how we can improve that.

sashko9807: Thanks for the explanation, by reading the front-end related issue, I was left with the impression that total amount of donations was supposed to be shown, rather than the total amount of users which have donated.
 
Aside from using raw SQL query to avoid prisma's limitation, the only improvement I could really think of is to order the  records by [null first](https://www.prisma.io/docs/concepts/components/prisma-client/filtering-and-sorting#sort-with-null-records-first-or-last)(if possible). That way `array.find`,  becomes O(1) in best-case scenario.

Also not sure if it is necessary to cover the unlikely usecase of `anonymousDonations` being undefined(there are no anonymous donations). In such scenarios the `donorsCount` will always be off by one.

Nnachevvv: Thanks for the suggestion. I have implemented it and now donations with billingName = "null" will appear always first , and in this way query is running in O(1).

So regarding the front-end issue , there is a small change. Instead of the number of active contributors, we decided to show the number of donors. So we will have both - total amount of donations and number of different donors.

igoychev (approved):
looks great!

PR #488, opened 2023-04-22
Manual import of unrecognized bank donations

BogoCvetkov:
Currently IRIS bank donations are imported automatically by a background worker. For the donations which have campaign codes that were not recognized, they are marked as `unrecognized`  or `importFailed` (if an error occurred during importing). There was no way for those bank donations to be processed manually by the administrators, so that the money are added to the correct camapaign vault.

This PR does exactly that. It adds a route that allows to set manually the correct payment code for unrecognized/failed bank donations.

## CHAGES
- Adds a new `matchedRef` field to the BankTransactions table. This field holds the value of the code (if any), that the system has recognized during importing(regardless if it belongs to a real campaign, it just matches that the code corresponds to the required structure). This was it's easier to debug how the system interprets the bank transaction and what is wrong

### New endpoints

`/bank-transaction/:id/edit-ref`


igoychev (approved):
all looks great and kudos for the tests!

PR #490, opened 2023-04-28
Non-english filenames in expenses fixed

slavcho:
Fix a bug with files uploaded as campaign expenses.
If the file is not in English it get scrabbled when sent via the form data upload.
So I encode that in base64 in the frontend and decode it here.

image


apps/api/src/expenses/expenses.service.ts
```diff
@@ -31,15 +26,12 @@ export class ExpensesService {
...
-          'File uploading: ',
-          expenseId,
-          file.originalname,
-          file.mimetype,
-          file.originalname,
-        )
+        //decode the name from base64
+        //multi-part form data does not support utf-8 filenames
+        const decodedFilename = Buffer.from(file.originalname, 'base64').toString('utf-8')
+        console.log('File uploading: ', expenseId, decodedFilename, file.mimetype)
```
igoychev, line 32: all good, just pls remove the console.log :)

igoychev (approved):

PR #499, opened 2023-06-02
Modification/notify admins on bank donations

BogoCvetkov:
Send notifications to platform administrators in 2 scenarios:

 - When an unrecognized bank donation has been detected
 - When less than 5 days are left to Bank Transaction Access Consent Expiration

In the first case all the unrecognized transaction ( from the ones that are currently being imported) are send as a table in an email. If an unrecognized donation was already sent in a previous email it will be skipped.

I've also done some small refactoring to the tasks folder - to make it more organized. Now we have a separate  service that has the sole responsibility of initializing and scheduling all the defined background-task and then we have  separate service for the different tasks (which get imported in the `TasksInitializerService`) - currently we only have `IrisTasks` service - containing all tasks related to Iris-related operations.

## Testing

1. Try simulating importing campaign with non-existing  campaign payment code as bank reference -> email should be sent with it's info
2.  Try creating a consent for less than 5 days -> an email should be sent

2 New Env Variables have been introduced

 [x] `ADMIN_NOTIFICATION_GROUP` = the group that will receive admin notifications
 [x] `CHECK_IRIS_CONSENT_TASK_HOUR` = At what hour of the day (each day), the check for the expiring bank consent should run


igoychev (approved):
great code and important improvement for the notifications!

PR #503, opened 2023-06-14
src/donations: Expand user's donations query to check for billingEmail

### LINKED ISSUE #500: Certificates for anonymous donations are not being received.
1. Sign in.
2. Choose a campaign, click 'Donate' button.
3. Choose payment method and donation sum.
4. Choose 'Donate anonymously' tab. 
5. Fill in email and complete the donation.
6. Go to Profile -> Donations.

The anonymous donation is not saved in the profile.

The donation has to be saved in the user's profile and certificate has to be received.

sashko9807:
Currently anonnymous donations are not included in user's profile. Solve this by making necessary queries look for match in either keycloakId, or billingEmail.

Note:
This solution would only work if the logged in user has, either left the email field empty, or has entered the same email as the one belonging to the account.

Closes #500 

---|---|---|---

### DIFF
--- a/apps/api/src/donations/donations.controller.ts
+++ b/apps/api/src/donations/donations.controller.ts
@@ -10,6 +10,9 @@ import {
   Query,
   Logger,
   Res,
+  Inject,
+  forwardRef,
+  NotFoundException,
 } from '@nestjs/common'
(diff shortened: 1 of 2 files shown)


igoychev (approved):
well done, simple and safe!

PR #509, opened 2023-06-30
Format admin notification emails

BogoCvetkov:
- Format the donation amount in unrecognized bank donation emails + currency
- Make the width of the emails for unrecognized donations/expiring consent bigger - so that it's easier to read
- Fix the title of the expiring bank consent email


igoychev (approved):
perfect!

PR #511, opened 2023-07-01
Integrate the Cache-manager to cache the most common APIs

slavcho:
Some of the data we render does not need to be in real time.
We can use a cache with a short timeout to protect against high load scenarios.

Tested this with 500 parallel requests.

Here are some stats for donors count API.
Without cache: 20 req/sec
With cache: 1200 req/sec.


apps/api/src/donations/donations.service.ts
```diff
@@ -41,6 +42,9 @@ export class DonationsService {
     private exportService: ExportService,
   ) {}
 
+  private cache = new HotCache()
+  private cacheTimeout = 30 * 1000 // 30 seconds
```
imilchev, line 46: could you make that an env variable? It would allow us to fine tune the value more easily. I was even thinking we can cache it for 1 or 2 mins. If it's exposed as a setting we can tweak it and see how it works

slavcho: Good idea.

sashko9807: Doesn't sound very scalable. Lets say for example, later on we decide to cache the result from the query which fetches all active and completed campaigns. Since this result is not expected to be as regularly updated, as the platform statistics, a bigger expiration time could be set

apps/api/src/common/hotcache.ts
```diff
@@ -0,0 +1,22 @@
+class CacheObject {
```
sashko9807, line 1: I think, we could get away without this class, and do everything in the `HotCache class` instead.

apps/api/src/common/hotcache.ts
```diff
@@ -0,0 +1,22 @@
...
+    constructor(public value: any, public expiration: number) {}
+    isExpired() {
+        return this.expiration < Date.now()
+    }
+}
+
+export class HotCache {
+    private cache: { [key: string]: CacheObject } = {}
```
sashko9807, line 9: Since we are going to deal with many insertions/deletions it is better to use [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), instead of regular objects

apps/api/src/common/hotcache.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    get(key: string) {
```
sashko9807, line 11: `get` and `set` are reserved words. Not that it is an issue, but consider renaming those methods, to avoid any possible confusion.

apps/api/src/common/hotcache.ts
```diff
@@ -0,0 +1,22 @@
...
+        const cacheObject = this.cache[key]
+        if (cacheObject && !cacheObject.isExpired()) {
```
sashko9807, line 13: We might want a separate condition to check if cacheObject.isExpired(), and if it is we should also 
`delete cacheObject`

apps/api/src/common/hotcache.ts
slavcho: Actually we currently have only 3 records, so no more than 3 insertions are expected.
The value will be updated every 30 seconds, but no deleting is planned.

apps/api/src/common/hotcache.ts
slavcho: Javascript is not my strongest language, but I don't see an issue here.
I need structure that holds both the value and the update time.

apps/api/src/donations/donations.service.ts
slavcho: we now have a default ttl in the config, but each service and variable can define its own expiration time.

apps/api/src/common/hotcache.ts
slavcho: done.

apps/api/src/common/hotcache.ts
slavcho: Don't see a reason to delete anything as of now.
This would only complicate things imho.

apps/api/src/common/hotcache.ts
slavcho: Just read that it can be a security issue to use an object here. 
Map is safer for keys that can come from user input.
Changed it.

apps/api/src/common/hotcache.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    const cacheObject = this.cache.get(key)
+    if (cacheObject && !cacheObject.isExpired()) {
+      return cacheObject.value
+    }
+    return null
+  }
+
+  // make the ttl optional
+  store(key: string, value: any, ttl?: number) {
```
igoychev, line 25: value: any raises a linter error due to noExplicitAny. 
In this context the safer type is 'unknown' as we don't know the type of the stored value, however at the place where used the type is known so type-checking will work well

igoychev (commented):
looks great!

just please replace 'any' with 'unknown' to avoid the linter error and have better type-safety

apps/api/src/donations/donations.service.ts
```diff
@@ -42,8 +42,7 @@ export class DonationsService {
     private exportService: ExportService,
   ) {}
 
-  private cache = new HotCache()
-  private cacheTimeout = 30 * 1000 // 30 seconds
+  private cache = new HotCache(this.config.get<number>('hotcache.ttl') || 0, !!this.config.get<boolean>('hotcache.enabled'))
```
sashko9807, line 45: Is it possible to inject the ConfigService, inside the `HotCache` class and get the configs from there, instead of getting them on every creation of instance?

imilchev:
we briefly looked at this at a time https://docs.nestjs.com/techniques/caching perhaps we should just use that

slavcho:
Switched to the cache-manager library.

manifests/base/deployment.yaml
```diff
@@ -224,3 +224,5 @@ spec:
                 secretKeyRef:
                   name: paypal-secret
                   key: webhook-id
+            - name: CACHE_TTL
+              value: 30000
```
imilchev, line 228: this needs to be quoted otherwise it will result in an error 
```suggestion
              value: "30000"
```

apps/api/src/config/configuration.ts
```diff
@@ -48,4 +48,8 @@ export default () => ({
   tasks: {
     import_transactions: { interval: process.env.IMPORT_TRX_TASK_INTERVAL_MINUTES },
   },
+  hotcache: {
```
imilchev, line 51: is this still needed?

manifests/base/deployment.yaml
imilchev: can we make this 1 or 2 min? 30s is not a lot of time and for the things that we cache it's not a problem if they don't refresh for a while

apps/api/src/config/configuration.ts
slavcho: Removed it.

manifests/base/deployment.yaml
slavcho: Changed it to 1 minute.

igoychev (approved):
cool!

PR #517, opened 2023-07-17
auth: Synchronize keycloak/person data on login

sashko9807:
A slight doubt exists, that in some rare occasions, there is a mismatch between the data saved in the keycloak server, and the one saved in the Person table. This mismatch could cause some unexpected behavior and unpleasant user experience, as we use the keycloakId of the logged user, for a lot of queries.
The proposed solution makes it so on every login, we check if the  email saved in the keycloak server exists in the Person table as well. 
The person record is updated on the following use cases:
1. Person record not found. In that case a new Person is being created during the login phase.
2. Person is found, but the keycloakIds are different. In that case the person record is being updated with the keycloakId coming from the keycloak server.

#### No user found use case
1. Run `yarn studio` go to Person, and delete the record, which contains the data for the authenticated user.
2. Login.
3. Go back to Prisma studio, and refresh the Person table. A new person record should be created containing the data for the authenticated user(email, firstName, lastName, keycloakId).
4. Compare the keycloakId from person record, with the one from the keycloak server to see if they match.

#### Keycloak mismatch use case
1.  Run `yarn studio` go to Person, find the Person record which contains the data for the authenticated user. Go to `keycloakId` field and edit it at random.
2. Login
3. Go back to Prisma studio, and refresh the Person table. Find the record of the authenticated user, go to `keycloakId` field, and check if it has been updated.
4. Compare the keycloakId from person record, with the one from the keycloak server to see if they match.

### DIFF
--- a/apps/api/src/auth/auth.service.spec.ts
+++ b/apps/api/src/auth/auth.service.spec.ts
@@ -19,6 +19,7 @@ import { ProviderDto } from './dto/provider.dto'
 import { EmailService } from '../email/email.service'
 import { JwtService } from '@nestjs/jwt'
 import { TemplateService } from '../email/template.service'
+import { PrismaService } from '../prisma/prisma.service'
 
 jest.mock('@keycloak/keycloak-admin-client')
 
@@ -27,6 +28,25 @@ describe('AuthService', () => {
   let config: ConfigService
   let admin: KeycloakAdminClient
   let keycloak: KeycloakConnect.Keycloak
+
+  const person: Person = {
+    id: 'e43348aa-be33-4c12-80bf-2adfbf8736cd',
+    firstName: 'Admin',
+    lastName: 'Dev',
+    keycloakId: '123',
+    email: 'test@podkrepi.bg',
+    emailConfirmed: false,
+    phone: null,
+    company: null,
+    picture: null,
+    createdAt: new Date('2021-10-07T13:38:11.097Z'),
+    updatedAt: new Date('2021-10-07T13:38:11.097Z'),
+    newsletter: false,
+    address: null,
+    birthday: null,
+    personalNumber: null,
+    stripeCustomerId: null,
+  }
 
   beforeEach(async () => {
     const module: TestingModule = await Test.createTestingModule({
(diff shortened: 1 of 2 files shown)


apps/api/src/auth/auth.service.ts
```diff
@@ -145,6 +145,22 @@ export class AuthService {
       if (!grant.access_token?.token) {
         throw new InternalServerErrorException('CannotIssueTokenError')
       }
+      const user = await this.keycloak.grantManager.userInfo<string, KeycloakTokenParsed>(
+        grant.access_token.token as string,
+      )
+      const person = await this.prismaService.person.findUnique({ where: { email: user.email } })
+      if (!person || person.keycloakId !== user.sub) {
+        Logger.warn('No person found for the current keycloak user. Creating new one...')
+        await this.authenticateAdmin()
```
slavcho, line 154: I assume we need authenticate as admin for the "createPerson" call.
Does this leave the user with admin rights after the login?

apps/api/src/auth/auth.service.ts
```diff
@@ -145,6 +145,22 @@ export class AuthService {
...
+        const userData = await this.admin.users.findOne({ id: user.sub })
```
slavcho, line 155: What if there is no user? I know it is highly unlikely, but maybe we should throw an error...

apps/api/src/auth/auth.service.ts
sashko9807: No. This is needed to check, whether the server itself, can access the keycloak realm, with the credentials in the .env

apps/api/src/auth/auth.service.ts
sashko9807: If there is no user in the keycloak server, login will be unsuccessful.

apps/api/src/auth/auth.service.spec.ts
```diff
@@ -167,27 +187,93 @@ describe('AuthService', () => {
...
   describe('login', () => {
-    const email = 'someuser@example.com'
+    const email = person.email ?? ''
+    const firstName = person.firstName ?? ''
+    const lastName = person.lastName ?? ''
     const password = 's3cret'
+    const keycloakId = person.keycloakId ?? '123'
 
-    it('should call issueGrant', async () => {
+    it('should call issueGrant - without updating person', async () => {
```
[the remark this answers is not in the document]
sashko9807: `expect(admin.auth).not.toHaveBeenCalled()` <- authenticateAdmin() is called before updating the person. If admin.auth has not been called, neither is `createPerson(...)`.

sashko9807:
@slavcho This PR will probably be closed at some point, as it was found out, that the issue which made me create this PR, was caused by something else...

slavcho:
Well - it is a good check anyway.
We have the test and all. It might never be triggered, but I don't see any harm in it.

PR #520, opened 2023-07-21
src/donations: Order public donations by updatedAt field

sashko9807:
When automatically importing bank transfers they are all created at the same time. As we currently order public donations by createdAt field, and the query is ran on every page change, this may cause unexpected behavior such as repeating of some bank donations in different pages, or some donations not being shown.
Instead use the updatedAt property which represents the exact time, donation record was added to database, or the last time its status has been updated. This should give us good approximation time of the donation itself.

Closes Reported in discord

## Motivation and context
Fixes a bug

#### Donations orderBy:
1. Call the endpoint responsible for listing donations:
`API_URL/api/v1/donation/listPublic/?campaignId=campaignId` where campaignId, is the id of the campaign in local enviroment.
2. Public donations should be ordered by their updatedAt property

#### updateAt change on personId/billingEmail change
1. Go to admin panel->Donations, Click F12 go to Network, and find the API response for the listed donations. Expand the first record and see its updatedAt property
2. Change donor field of the first record, of listed donations.  The updatedAt property of the first record, should be the same as in the previous API response.
3. Repeat the same for the billingEmail field

### DIFF
--- a/apps/api/src/donations/donations.service.ts
+++ b/apps/api/src/donations/donations.service.ts
@@ -605,7 +605,9 @@ export class DonationsService {
         data: {
           status: status,
           personId:  updatePaymentDto.targetPersonId ? donorId : undefined,
-          billingEmail: updatePaymentDto.billingEmail ? billingEmail : undefined 
+          billingEmail: updatePaymentDto.billingEmail ? billingEmail : undefined,
+          //In case of personId or billingEmail change, take the last updatedAt property to prevent any changes to updatedAt property
+          updatedAt: updatePaymentDto.targetPersonId || updatePaymentDto.billingEmail ? currentDonation.updatedAt : undefined
         },
       })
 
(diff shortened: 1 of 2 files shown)


sashko9807:
Side note:
As the updatedAt field could potentially cause some unwanted side-effects, we should thoroughly, observe every scenario, in which it changes, before a merge(if it is merged).

apps/api/src/donations/donations.service.ts
```diff
@@ -604,8 +601,13 @@ export class DonationsService {
...
           status: status,
-          personId:  updatePaymentDto.targetPersonId ? donorId : undefined,
-          billingEmail: updatePaymentDto.billingEmail ? billingEmail : undefined 
+          personId: updatePaymentDto.targetPersonId ? donorId : undefined,
+          billingEmail: updatePaymentDto.billingEmail ? billingEmail : undefined,
+          //In case of personId or billingEmail change, take the last updatedAt property to prevent any changes to updatedAt property
+          updatedAt:
+            updatePaymentDto.targetPersonId || updatePaymentDto.billingEmail
+              ? currentDonation.updatedAt
+              : undefined,
```
igoychev, line 610: Well done with this smart check to preserve the original update!

igoychev (approved):

PR #521, opened 2023-07-26
User notifications - Part 1

BogoCvetkov:
Depend on this FE PR -> podkrepi-bg/frontend#1517

Added functionality to subscribe to marketing notifications in SendGrid. Two types of notifications would exist.
1. General marketing notifications (Platform realated news, New campaign has been published, Blog Posts etc),
2. Campaign Specific notifications - % of amount raised, campaign news, campaign completed etc.a

 - So basically this PR takes care of generating mailing lists in SendGrid and managing subscriptions to them(adding emails). The actual sending would be implemented in the next PR. Visual example of how those lists look like in SendGrid:

image

- The `All Marketing Emails` List is where all emails that gave their consent(registered and not) are placed
- Additionaly if an email subscribes to a particular campaign it will also be added to the list for that camapign
 
NOTE: This is the First part of the implementation. The next one will add:

1. Ability to un-subscribe - logged in or not
2. Ability to manage notification consent and saparate campaign notification subscriptions of logged users
4. The actual sending of emails to the different mailing groups (General and Campaign based)

## How it works
1. In the register route a new flag is accepted `{newsletter:boolean}` which marks if the user has accepted to receive marketing notifications. A checkBox has been added on the FE as well.
  - If the flag is true the user consent is saved in the DB and the email itself is added ot the `All Marketing Emails` List to receive notifications
 2. In the route for updating campaigns -  logic has been added so that the mailing lists are created automatically in Sendgrid for the updated campaign. The logic is as follows:
  - If no list exists for the campaign and it's state is updated to active - an API call is made to SendGrid to create the lists with the campaign name and also another record in a new table in the DB is created that tracks the mailing list SendGrid Id for that camapaign
  - If the campaign is active and already has an email list and If the campaign title is updated and api call is send to SendGRid to update the list with the new name
  3. A new route is created in the campaign controller - `/campaign/:id/subscribe`. It is `PUBLIC`, so that all interested visitors can subscribe to it. You can check the FE PR to see how it should work. The consent for non-registered emails are stored in a new DB table.  NOTE: currently it directly adds the provided email to the `All Marketing Emails` List + the mailing list for that campaign. In the next PR this shall get modified, so that we will send an email that would contain the actual verification link. This way we would filter out some spam and fake emails.
 4. Wrote tests for the new functionality

NOTE: The Service for managing notifications is implemented trough an `INTERFACE`, the idea is that it should be easier to switch to another marketing provider if we decide that SendGrid is not OK for us.

### New endpoints

`/campaign/:id/subscribe`

MARKETING_LIST_ID

- [ ] `MARKETING_LIST_ID`: the Sendgrid list where all emails that gave consent would be stored - a.k.a All Marketing Emails or whatever it's called

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -69,6 +69,7 @@ SENDGRID_API_KEY=sendgrid-key
 SENDGRID_SENDER_EMAIL=info@podkrepi.bg
 SENDGRID_INTERNAL_EMAIL=dev@podkrepi.bg
 SENDGRID_CONTACTS_URL=/v3/marketing/contacts
+MARKETING_LIST_ID=6add1a52-f74e-4c14-af56-ec7e1d2318f0
 
 ## Stripe ##
 ############
--- a/apps/api/src/auth/auth.service.ts
+++ b/apps/api/src/auth/auth.service.ts
@@ -30,6 +30,8 @@ import { JwtService } from '@nestjs/jwt'
 import { EmailService } from '../email/email.service'
 import { ForgottenPasswordMailDto } from '../email/template.interface'
 import { NewPasswordDto } from './dto/recovery-password.dto'
+import { NotificationsInterface } from '../notifications/notifications.interface'
+import { SendGridParams } from '../notifications/notifications.sendgrid.types'
 
 type ErrorResponse = { error: string; data: unknown }
 type KeycloakErrorResponse = { error: string; error_description: string }
@@ -60,6 +62,7 @@ export class AuthService {
     private jwtService: JwtService,
     private sendEmail: EmailService,
     @Inject(KEYCLOAK_INSTANCE) private keycloak: KeycloakConnect.Keycloak,
+    private readonly marketingNotificationsService: NotificationsInterface<SendGridParams>,
   ) {}
 
   async issueGrant(email: string, password: string): Promise<KeycloakConnect.Grant> {
@@ -160,12 +163,13 @@ export class AuthService {
   }
 
   async createUser(registerDto: RegisterDto): Promise<Person | ErrorResponse> {
+    let person: Person
     try {
       await this.authenticateAdmin()
       // Create user in Keycloak
       const user = await this.createKeycloakUser(registerDto, false)
       // Insert or connect person in app db
-      return await this.createPerson(registerDto, user.id)
+      person = await this.createPerson(registerDto, user.id)
     } catch (error) {
       const response = {
         error: error.message,
@@ -174,6 +178,28 @@ export class AuthService {
       Logger.error(response)
       return response
     }
+
+    // Add to marketing platform
+    if (registerDto.newsletter)
+      try {
+        // Add email to general marketing notifications list
+        const mainList = this.config.get('sendgrid.marketingListId')
+        if (mainList)
+          await this.marketingNotificationsService.addContactsToList({
+            contacts: [
+              {
+                email: person.email,
+                first_name: person?.firstName || '',
+                last_name: person?.lastName || '',
+              },
+            ],
+            list_ids: [mainList],
+          })
+      } catch (e) {
+        Logger.error('Failed to subscribe email', e)
+      }
+
+    return person
   }
 
   private async authenticateAdmin() {
@@ -228,6 +254,7 @@ export class AuthService {
         email: registerDto.email,
         firstName: registerDto.firstName,
         lastName: registerDto.lastName,
+        newsletter: registerDto.newsletter ? registerDto.newsletter : false,
       },
       // Store keycloakId to the person with same email
       update: { keycloakId },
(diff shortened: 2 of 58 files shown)


apps/api/src/campaign/campaign.service.ts
```diff
@@ -713,8 +720,69 @@ export class CampaignService {
         },
       })
     }
+    // Create a notification list for this campaign when updated to active
+    if (
+      campaign?.state !== CampaignState.active &&
```
igoychev, line 725: please add CampaignState.approved too

apps/api/src/campaign/campaign.service.ts
```diff
@@ -713,8 +720,69 @@ export class CampaignService {
...
+      id: list.id,
+    })
+    // update in DB
+    await this.prisma.notificationList.update({
+      where: { id: list.id },
+      data: {
+        name: updated.title,
+      },
+    })
   }
```
igoychev, line 786: These two methods look like specific to the NotificationService - my suggestion is to move them there

apps/api/src/auth/auth.service.ts
```diff
@@ -190,6 +194,28 @@ export class AuthService {
...
+                last_name: person?.lastName || '',
+              },
+            ],
+            list_ids: [mainList],
+          })
+      } catch (e) {
+        Logger.error('Failed to subscribe email', e)
+      }
+
+    return person
```
igoychev, line 218: this code seems like specific to the NotificationService - let's move it there, because the auth service shouldn't know about list configs

.env.example
```diff
@@ -69,6 +69,7 @@ SENDGRID_API_KEY=sendgrid-key
 SENDGRID_SENDER_EMAIL=info@podkrepi.bg
 SENDGRID_INTERNAL_EMAIL=dev@podkrepi.bg
 SENDGRID_CONTACTS_URL=/v3/marketing/contacts
+MARKETING_LIST_ID=6add1a52-f74e-4c14-af56-ec7e1d2318f0
```
igoychev, line 72: is this the main newsletter list Id for all news from Podkrepi.bg? 
If so then let's call it PODKREPI_NEWSLETTER_LIST_ID

apps/api/src/notifications/notifications.interface.ts
```diff
@@ -0,0 +1,15 @@
+type NotificationsInterfaceParams = {
+  CreateListParams: any
+  UpdateListParams: any
+  DeleteListParams: any
+  AddToListParams: any
+  RemoveFromListParams: any
+}
```
igoychev, line 7: please replace any with unknown for typesafety and the linter to be happy :)

apps/api/src/notifications/notifications.sendgrid.service.ts
```diff
@@ -0,0 +1,84 @@
+import { Injectable, Logger } from '@nestjs/common'
+import { ConfigService } from '@nestjs/config'
+import sgClient from '@sendgrid/client'
+import { NotificationsInterface } from './notifications.interface'
+import { SendGridParams } from './notifications.sendgrid.types'
+import { ClientRequest } from '@sendgrid/client/src/request'
+
+@Injectable()
+export class SendGridNotificationsService implements NotificationsInterface<SendGridParams> {
```
igoychev, line 9: Great architectural choice for implementing it with Interface

apps/api/src/person/person.service.ts
```diff
@@ -38,6 +38,10 @@ export class PersonService {
     return await this.prisma.person.findFirst({ where: { id } })
   }
 
+  async findByEmail(email: string) {
+    return await this.prisma.person.findFirst({ where: { email } })
```
igoychev, line 42: use findUnique here as the email should be unique

apps/api/src/person/person.service.ts
```diff
@@ -71,4 +75,20 @@ export class PersonService {
...
+        email,
+        consent,
+      },
+      // Update consent
+      update: { consent },
+      where: {
+        email,
+      },
+    })
+  }
```
igoychev, line 93: for the unregistered persons, I would expect this function to be in the NotificationsService because the PersonService is for registered users

igoychev (changes requested):
Wow, great implementation! 
I have just a few cosmetic and encapsulation comments otherwise the logic looks solid!

apps/api/src/campaign/campaign.controller.ts
```diff
@@ -167,6 +173,83 @@ export class CampaignController {
       )
   }
 
+  @Post(':id/subscribe')
+  @Public()
+  async subscribeToCampaignNotifications(
+    @Param('id') id: string,
+    @Body() data: CampaignSubscribeDto,
+  ) {
```
igoychev, line 181: this function looks like quite long for to be in a controller - is it possible to break down and move the specific parts to the specific service for better encapsulation?

PR #527, opened 2023-08-03
Create a new internal campaing for members and their fees.

slavcho:
We already support subscriptions on the platform.
Now we can use those subscriptions to pay the annual membership fees.

The campaign that is created has an approved status.
So it is not public, but can be accessed via this link:

https://podkrepi.bg/campaigns/podkrepi-membership

Closes https://github.com/podkrepi-bg/frontend/issues/1530 

### Steps to test
Open the link:
https://podkrepi.bg/campaigns/podkrepi-membership

Then click to donate and select the subscription checkbox.
Although it says "monthly", the donation will be annual in Stripe:

image

Currently we select a random coordinator and a beneficiary, but this will be changed after deploying on production.

### DIFF
--- a/apps/api/src/campaign/campaign.service.ts
+++ b/apps/api/src/campaign/campaign.service.ts
@@ -791,4 +791,12 @@ export class CampaignService {
 
     return articles
   }
+
+  async isMembershipCampaign(campaignTypeId: string): Promise<boolean> {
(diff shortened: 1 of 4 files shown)


slavcho:
Trying to fix the migration problem. Don't review just yet.

igoychev (approved):
cool!

PR #528, opened 2023-08-05
Use wishes campaignId param

tongo-angelov:
Use the campaignId param as to not need to send requests to ...donation-wish/list/{campaignId}?campaignId={campaignId}...


igoychev (approved):
good catch - well done with the optimization

Issue #529, opened 2023-08-06
[donations] check for campaign completion should happen inside the donation db transaction scope 


igoychev:
### Describe the Bug

@BogoCvetkov has reported a bug, where the check for reaching the total amount is not inside the same prisma db transaction scope. This causes the campaign to be marked as completed one donation after the reached amount.

### To Reproduce

Add the logic for checking the total amount to be called inside the prisma transaction.

### Expected Behavior

Campaign should be marked as completed with the first donation that reaches the total amount.

PR #536, opened 2023-08-15
fix campaign completion transaction

quantum-grit:
## Motivation and context

Closes https://github.com/podkrepi-bg/api/issues/529

Added db transaction when Creating and updatingDonation to succeeded that encapsulates the completion of campaign when targetAmount is reached.

Reviewing by the final diff is going to be hard, so I recommend reviewing commits one by one.

All changes:
- deleted manual donation create method as it is not allowed
- moved incrementVaultAmount inside updateDonationPayment and added the transaction wrapper
- fixed existing tests
- fixed tests with the expected update vault
- linter cleanup
- simplified updateDonationPayment by extracting its parts into smaller functions

## Testing
yarn test


apps/api/src/campaign/campaign.service.ts
```diff
@@ -450,95 +450,71 @@ export class CampaignService {
...
-            billingEmail: paymentData.billingEmail,
-            person: paymentData.personId ? { connect: { id: paymentData.personId } } : {},
-          },
-          select: donationNotificationSelect,
-        })
-        if (newDonationStatus === DonationStatus.succeeded) {
-          this.notificationService.sendNotification('successfulDonation', donation)
+      //For successful donations we will also need to link them to user if not marked as anonymous
+      if (donationId && newDonationStatus === DonationStatus.succeeded) {
+        if (metadata?.isAnonymous != 'true') {
```
BogoCvetkov, line 491: !==   - just to be consistent and sure that equality will work as expected

apps/api/src/campaign/campaign.service.ts
```diff
@@ -573,26 +554,88 @@ export class CampaignService {
...
-              connect: {
-                email: paymentData.billingEmail,
-              },
-            },
-          },
-        })
+    /**
+     * Create or connect campaign vault
+     */
+    const vault = await this.prisma.vault.findFirst({ where: { campaignId: campaign.id } })
```
BogoCvetkov, line 575: shouldn't we be using the `tx` object, just to be sure that we are always seeing the state of each record as it is in the transaction and prevent any inconsistencies

apps/api/src/campaign/campaign.service.ts
```diff
@@ -553,12 +529,17 @@ export class CampaignService {
...
+        //if donation is switching to successful, increment the vault amount and send notification
         if (newDonationStatus === DonationStatus.succeeded) {
+          await this.vaultService.incrementVaultAmount(donation.targetVaultId, donation.amount, tx)
           this.notificationService.sendNotification('successfulDonation', {
             ...updatedDonation,
-            person: donation.person,
+            person: updatedDonation.person,
           })
         }
+
```
BogoCvetkov, line 541: should we add the `
 if (metadata?.isAnonymous != 'true') {
          where: { id: donation.id  ......
` logic for associating an email with a donation when updating

apps/api/src/campaign/campaign.service.ts
```diff
@@ -553,12 +529,17 @@ export class CampaignService {
           },
           select: donationNotificationSelect,
         })
+
+        //if donation is switching to successful, increment the vault amount and send notification
         if (newDonationStatus === DonationStatus.succeeded) {
+          await this.vaultService.incrementVaultAmount(donation.targetVaultId, donation.amount, tx)
```
BogoCvetkov, line 535: Would there be a problem if this function get's called a second time (another update) and this code block get's re-executed

apps/api/src/donations/events/stripe-payment.service.spec.ts
```diff
@@ -221,9 +223,14 @@ describe('StripePaymentService', () => {
       personId: 'donation-person',
     })
 
-    const mockedDonateToCampaign = jest
-      .spyOn(campaignService, 'donateToCampaign')
-      .mockName('donateToCampaign')
+    jest.spyOn(prismaMock, '$transaction').mockImplementation((callback) => callback(prismaMock))
+    const mockedupdateDonationPayment = jest
```
slavcho, line 227:
```suggestion
    const mockedUpdateDonationPayment = jest
```

slavcho (approved):
Looks good to me.

apps/api/src/campaign/campaign.service.ts
quantum-grit: yes, thanks! 
this is important as the donation creation could fail, if another donation creates the same vault in meantime

apps/api/src/campaign/campaign.service.ts
quantum-grit: the incrementVaultAmount is executed only on new successful donation or when a waiting donation is turned into successful, so no, there is no way this could be called again on the same donation

apps/api/src/campaign/campaign.service.ts
quantum-grit: done

PR #539, opened 2023-08-22
src/campaign-file: Set Cache-Control header

sashko9807:
This would allow for images to be cached on the client, and thus result in faster load on client, as well as less calls to the server being made.  
Age of cache  is set to 6 months. 

Test:
1. Open Browser's devtools
2. Go to network tab
3. Refresh page
4. Campaign files should be loaded through disk cache.

### DIFF
--- a/apps/api/src/campaign-file/campaign-file.controller.ts
+++ b/apps/api/src/campaign-file/campaign-file.controller.ts
@@ -12,7 +12,6 @@ import {
   Body,
   Inject,
   forwardRef,
-  Header,
 } from '@nestjs/common'
 import { FilesInterceptor } from '@nestjs/platform-express'
 import { UseInterceptors, UploadedFiles } from '@nestjs/common'
@@ -25,6 +24,7 @@ import { CampaignFileService } from './campaign-file.service'
 import { CampaignService } from '../campaign/campaign.service'
 import { KeycloakTokenParsed, isAdmin } from '../auth/keycloak'
 import { ApiTags } from '@nestjs/swagger';
+import { CampaignFileRole } from '@prisma/client'
 
 @ApiTags('campaign-file')
 @Controller('campaign-file')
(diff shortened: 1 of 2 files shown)


slavcho (commented):
Ha, I was preparing something similar with the Expires header.
This seems better...

Why not include all the campaign Photos in the cached images?

  campaignPhoto: 'campaignPhoto',
  profilePhoto: 'profilePhoto',
  campaignListPhoto: 'campaignListPhoto',
  beneficiaryPhoto: 'beneficiaryPhoto',
  organizerPhoto: 'organizerPhoto',

sashko9807:
> Ha, I was preparing something similar with the Expires header. This seems better however...
> 
> Why not include all the campaign Photos in the cached images?
> 
> campaignPhoto: 'campaignPhoto', profilePhoto: 'profilePhoto', campaignListPhoto: 'campaignListPhoto', beneficiaryPhoto: 'beneficiaryPhoto', organizerPhoto: 'organizerPhoto',

I wanted to prevent accumulating a lot of memory inside the browser's cache by caching only the data that is initially shown to user, as well as the data which is not susceptible to frequent changes.

Caching other images should probably be done on nginx/cloudfare level.

slavcho:
Caching a few more images in the browser is not going to take a lot of memory.
The conversation started with this issue: https://github.com/podkrepi-bg/frontend/issues/1549
Here the assumption is that it happens with very bad connections. 

There is no point in caching invoices and PDFs - I agree.
But I think we should cache all the images.

sashko9807:
> Caching a few more images in the browser is not going to take a lot of memory. The conversation started with this issue: [podkrepi-bg/frontend#1549](https://github.com/podkrepi-bg/frontend/issues/1549) Here the assumption is that it happens with very bad connections.
> 
> There is no point in caching invoices and PDFs - I agree. But I think we should cache all the images.

MM yea. Seems like the  `<Image>` component from next js, also caches the images locally, so it should be fine.

PR #549, opened 2023-09-15
Add sorting for bank transactions

dphilipov:
Related to [#1588](https://github.com/podkrepi-bg/frontend/issues/1588)

## Motivation and context

* My initial solution was to sort the data on the front-end, but since the data is paginated, this sorts only the current page. This PR makes sure the sort details reach the query, thus sorting all bank transactions.


apps/api/src/bank-transactions/bank-transactions.service.ts
```diff
@@ -65,6 +67,11 @@ export class BankTransactionsService {
       },
       skip: pageIndex && pageSize ? pageIndex * pageSize : undefined,
       take: pageSize ? pageSize : undefined,
+      orderBy: [
+        {
+          [sortBy as string]: sortOrder,
+        },
+      ],
```
igoychev, line 74: might be good to set a default sorting if the param is not sent by frontend, something like:
orderBy: [sortBy ? { [sortBy]: sortOrder ? sortOrder : 'desc' } : { transactionDate: 'desc' }],

igoychev (commented):
looks great!
just please check if the frontend is always passing the sortBy and sortOrder or add some defaults like in the comment below

apps/api/src/bank-transactions/bank-transactions.service.ts
dphilipov: Implemented :+1:

igoychev (approved):
wonderful!

PR #550, opened 2023-09-17
Handle payment_intent.payment_failed when card is not accepted

### LINKED ISSUE #548: [Stripe] handle payment_intent.payment_failed

Donations

When payment is initiated with payment_intent.created event and the card is not accepted by some reason in Stripe, it sends payment_intent.payment_failed event which we currently don't handle. 

This results in donations staying in status waiting, however the closest correct status should be 'declined'.

Check how @StripeWebhookHandler('payment_intent.canceled') is implemented and add implementation for payment_intent.payment_failed

Testing of stripe events is described here: https://github.com/podkrepi-bg/api/blob/master/TESTING.md

_No response_

Nnachevvv:
When payment is initiated with payment_intent.created event and the card is not accepted by some reason in Stripe, it sends payment_intent.payment_failed event which we currently don't handle.

This results in donations staying in status waiting, however the closest correct status should be 'declined'.

This PR fixes that

Closes #548 

### Steps to test
1. I followed the document which we have in order to setup stripe:
https://github.com/podkrepi-bg/api/blob/master/TESTING.md

2. Executed stripe trigger:
```bash
stripe trigger payment_intent.payment_failed --override payment_intent:metadata.campaignId=a57c8d72-39e0-45e8-ab59-ea84f371a35a --override payment_intent:currency=BGN
```

Donations that have been declined before executing the command above: 
Screenshot 2023-09-18 at 0 06 11

Donations after executing the command above: 
Screenshot 2023-09-18 at 0 07 30


igoychev (approved):
looks great, and also kudos for adding the test!

PR #552, opened 2023-09-19
Return full beneficiary info

tongo-angelov:
Return all the required info for the admin/users edit form as to not call second endpoint


igoychev (approved):
looks great!

PR #556, opened 2023-10-01
[Stripe] Implement Refund event listener in Stripe webhook

### LINKED ISSUE #518: [Stripe] Implement Refund event listener in Stripe webhook

Donations

We need to handle Stripe refund event, so that in case of Refund request, the refunded amounts to be properly substracted. 

The refund itself is triggered by a billing admin in Stripe Admin UI and then Stripe will send us a redund.created event: https://stripe.com/docs/api/refunds/object

Note: there is a frontend issue also for implementing a Refund button in the AdminUI: https://github.com/podkrepi-bg/frontend/issues/1585

When handling refund.created please check:
 - if it is the only event that confirms that the refund was successful. 
 - if refunded amount will return the whole amount or the the amount without taxes - we may need to send email notification to the billing_email of the donation

The amount of the donation need to be substracted from the vault in similar way that it is added to it when successful donation is received. 

_No response_

Nnachevvv:
This PR ensures we will handle stripe refund event, so that in case of Refund request, the refunded amounts to be properly substracted.

The refund itself is  by calling **/api/v1/donation/refund-stripe-paymet/<extPaymentId>** endpoint or from billing Stripe Admin UI. After refund request is received Stripe sends a charge.refunded event: https://stripe.com/docs/api/refunds/object

After we receive this event, we will:
- subtract the value from the vault
- send an email to the end user
- change the donation status to refund

Relates to https://github.com/podkrepi-bg/frontend/pull/1635
Closes #518 and https://github.com/podkrepi-bg/frontend/issues/1585

Test:

Created a donation:
Screenshot 2023-10-01 at 17 27 32

After refund button is clicked - donation is not present in the campaign tab:

Screenshot 2023-10-28 at 23 02 24

Screenshot 2023-10-28 at 23 02 40

Screenshot 2023-10-01 at 17 28 21

End user will receive the following mail:

Screenshot 2023-10-01 at 17 34 34


igoychev (changes requested):
all great!
the logic is implemented on the right places and I especially like how you made the refactoring around Increment/Decrement to preserve readability and correctness.

my only request for change is the text in the email as per the comment

igoychev (approved):

PR #557, opened 2023-10-01
fixed skipped bank transactions

quantum-grit:
## Motivation and context
It was reported that some bank transactions are missing. 
To fix, now transactions are requested with dateTo = dateFrom+1 to ensure all transactions are returned from bank API
Also, removed checking for existing records by count as it may result in skipped transactions


igoychev (approved):
looks great

PR #564, opened 2023-10-19
[2/2] Affiliate program Integration v1

### LINKED ISSUE #512: [AffiliateAPI] Create API for Affiliates to submit donations

_No response_

Some companies has expressed interest in integrating donations within their systems and want to submit donations automatically as part of their application user experiences.

To enable this we need to provide:
 - API endpoint for submitting donations with AffiliateId to assocate donations with
 - The Affiliate company wants to see donations immediately, however the bank transfers will be made at bulk once per month
 - The donations should be reflected immediately on the web site, however they need to be marked with status "guaranteed" and when the monthly transfer comes and total amount is confirmed they will be marked as successful
 - This works similarly as card amounts are first authorized and then completed.

To implement the requirements the backend needs to support the following:
- [ ] Public API endpoint with POST request for accepting the affiliate donations. Proposed url: /donations/affiliate/:key/donate where key is a secret key shared only between Podkrepi and the Affiliate
- [ ] Public API endpoint to GET affiliated donations by affiliate - this could be parameter to existing donation list API
- [ ] Private API endpoint for billing admins to register new Affiliate and return newly generated API key
- [ ] Private API endpoint for billing admins to create new API key in case it was compromised

For consolidation purposes:
- [ ] The current automated bank transaction listener needs to be extended to recognize the Affiliate in the incoming bank transfers by the sender name. The code then needs match and confirm the "guaranteed" amounts into "successful" status. The Affiliate bank sender account name needs to be filled when the Affiliate is registered.

reusing the current donation API endpoint is not desired due to other security concerns. We want separate workflows for Affiliates to not accidentally reduce the security or change the donation statuses.

sashko9807:
Closes #512 

## Motivation and context

Opens donation API for 3rd parties

#### Endpoint

```http
POST http://localhost:5010/api/v1/register
```

#### Body
```json
{
    "type": "corporate",
    "companyName": "Association Podkrepi BG",
    "companyNumber": "206398075",
    "email": "john.doe@podkrepi.bg",
    "password": "$ecurePa33",
    "firstName": "John",
    "lastName": "Doe"
}
```
### 2.Activate new profile from keycloak
`https://localhost:8180` -> `login` -> `webapp realm` -> `Users` -> `View all users` -> `Click on profile with john.doe@podkrepi.bg` -> `User enabled` -> `Save`

### 3. Join affiliate program
1. Login with the email and password from above, through `/login` endpoint
2. Copy the returned accessToken.
3. Send request to the below endpoint, with the accessToken copied from above set in the Authorization header.(If prisma studio is started, a new affiliate should be seen with status `pending`
```http
POST http://localhost:5010/api/v1/affiliate/join
```
### 4. Activating affiliate
1. Login with admin user and copy the accessToken, and attach it to the Authorization header for the request below.
2. Send request to the below endpoint, where affiliateId, is the id of the requested affiliate

#### Endpoint
```http
PATCH http://localhost:5010/api/v1/affiliate/:affiliateId/status
```
#### Body
```json
{
    "newStatus": "active"
}
```
3. When viewing Affiliate table in Prisma studio, the affiliate's status should be active and affiliateCode should be with value with structure `af_xxxxx`.

#### Endpoint
```http
POST http://localhost:5010/api/v1/affiliate/:affiliateCode/donation
```
#### Body
```json
{
    "amount": 50000,
    "billingName": "John doe",
    "campaignId": ":campaignId",
    "isAnonymous": false
}
```
#### Endpoint
```http
PATCH http://localhost:5010/api/v1/affiliate/:affiliateCode/donations/:donationId/cancel
```

#### Endpoint
```http
POST http://localhost:5010/api/v1/bank-transaction/iris-transaction-test
```
#### Body
```json
{
    "irisIbanAccountInfo" : {
        "iban": "BG80BNBG96611020345678",
        "bankName": "UniCredit"
    },
    "irisTransactionInfo": [{
        "transactionId": "Booked_5954782144_70123543493054963FTRO23073A58G01C2023345440_20230320",
        "bookingDate": "2023-03-17",
        "creditorAccount": {
            "iban": "BG66UNCR70001524349032"
        },
        "creditorName": "СДРУЖЕНИЕ ПОДКРЕПИ БГ",
        "debtorAccount": {
            "iban": "BG66UNCR70001524349032"
        },
        "debtorName": "JOHN DOE",
        "remittanceInformationUnstructured": ":affiliateCode",
        "transactionAmount": {
            "amount": 5000,
            "currency": "BGN"
        },
        "exchangeRate": null,
        "valueDate": "2023-03-14",
        "creditDebitIndicator": "CREDIT"        
    }   
    ]
}
```

#### Affiliate
`/affiliate/join` - Endpoint through which corporate user can request to join affiliate program
`/affiliate/:affiliateId/status-update` - **Admin only** Endpoint through which affiliate's status can be updated (e.g `active`, `rejected` etc.)
`/affiliate/:affiliateCode/donation`  - Endpoint through which **active** affiliates can make guaranteed donations
`/affiliate/:affiliateCode/donations/:donationId/cancel` - Endpoint through which **active** affiliates can cancel guaranteed donation.

#### Bank Transactions
`/bank-transaction/iris-transaction-test` - **ADMIN/AFFILIATES ONLY** **DEV ENVIROMENT ONLY**  Endpoint through which an IRIS response can be simulated


BogoCvetkov:
Currently the PR says there are 250 files changed, where probably 80% of them were just de-formated -> something that happens when new migrations are generated when running `npm prisma migrate dev`. Maybe it's a good idea to run `yarn format` to re-format the files as they were before that, so that they don't appear here as changes and we can focus on the files that contain the new logic

sashko9807:
> Currently the PR says there are 250 files changed, where probably 80% of them were just de-formated -> something that happens when new migrations are generated when running `npm prisma migrate dev`. Maybe it's a good idea to run `yarn format` to re-format the files as they were before that, so that they don't appear here as changes and we can focus on the files that contain the new logic

Done. Sorry for that.

schema.prisma
```diff
@@ -86,14 +86,28 @@ model Company {
   legalPersonName String?       @map("legal_person_name")
   countryCode     String?       @map("country_code") @db.Citext
   cityId          String?       @map("city_id") @db.Uuid
+  personId        String?       @unique @map("person_id") @db.Uuid
```
igoychev, line 89: With this change we are saying that one or many companies can be associated with one person and this could be quite restrictive.
It looks more natural for the connection to be many people to be associated with one company. So the suggestion is to put companyId in the Person table.

apps/api/src/auth/auth.service.ts
```diff
@@ -225,13 +231,17 @@ export class AuthService {
...
+    verifyEmail: boolean,
+    activeProfile = true,
+  ) {
     return await this.admin.users.create({
       username: registerDto.email,
       email: registerDto.email,
       firstName: registerDto.firstName,
       lastName: registerDto.lastName,
-      enabled: true,
+      enabled: activeProfile,
```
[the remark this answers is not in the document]
sashko9807: The idea is to manually active corporate profiles after verification, to avoid situations such as Ivan Ivanov who has nothing to do with Google, to create corporate profile for Google and make donations using Google's name.
Whether this is actually needed I am unsure. 
How we will verify the corporate profile I'm also unsure.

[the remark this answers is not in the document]
igoychev: oh, yes! I remember now - thanks and this is indeed ok!

apps/api/src/affiliate/affiliate.controller.ts
```diff
@@ -0,0 +1,114 @@
...
+    @Body() donation: CreateAffiliateDonation,
+  ) {
+    const affiliate = await this.affiliateService.findOneByCode(affiliateCode)
+    if (!affiliate || !affiliate.company || !affiliate.company.person)
+      throw new NotFoundException('Affiliate not found')
+
+    const affiliateDonationDto: CreateAffiliateDonation = {
+      ...donation,
+      affiliateId: affiliate.id,
+      personId: donation.isAnonymous ? null : affiliate.company.person.id,
```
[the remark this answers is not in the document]
sashko9807: The `affiliate.company.person.id` , represents the company, thus any donations connected to that field, should be considered to be done by the company, rather than the individual person behind the profile, be it affiliate donation or regular donation.

The `personId` of the (end end user), doesn't exist in our database records. We could expand the request body to include(personEmail, personName), and create a new person, but then donation certificate will be issued for that individual person, rather than for the company, which is actually making the transaction. To avoid that we could also expand the affiliate API, to allow the affiliate admin to create more corporate profiles(which don't require manual activation), but then the keycloak should be configured to allow only for the **affiliate administrator** to be able to create such profiles, and it kind of becomes overcomplicated.

The idea was to use the `billingName` field to mark the exact person who made the donation(which I seem to forgot to put inside the object). 
Seems like I also forgot to include the field `extCustomerId` which would allow the affiliate to internally track the profile behind the donation.

Overall marking the actual person behind the affiliate donation is kind of hard. Even if we use `billingName` field to get that person, we need to add additional complexity to the `listPublic` endpoint, to transform the donations results, in a way that `billingName` is only included in the response, if the donation is affiliate, and it isn't anonymous.

If we just include `billingNamel` in the listPublic response, then no donation would actually be anonymous. 

I will include the  the forgotten `billingName` and  `extCustomerId` fields in the createDonationDto, but perhaps the whole process of  getting the exact person behind this affiliate donation could be done  later?

apps/api/src/donations/donations.service.ts
```diff
@@ -414,6 +426,18 @@ export class DonationsService {
     }
   }
 
+  async getAffiliateDonationById(donationId: string, affiliateCode: string) {
+    try {
+      const donation = await this.prisma.donation.findFirstOrThrow({
+        where: { id: donationId, affiliate: { affiliateCode: affiliateCode } },
+      })
+      return donation
```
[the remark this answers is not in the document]
sashko9807: There is still a very small theoretical possibility, of having one affiliate cancel the donation of another one. Putting affiliateCode inside the query ensures that the  donation , is made by the affiliate who desires to cancel it.

apps/api/src/tasks/bank-import/import-transactions.task.ts
```diff
@@ -377,6 +455,26 @@ export class IrisTasks {
...
+    for (const donation of affiliate.donations) {
+      if (trx.amount - totalDonated < donation.amount) continue
+      await this.donationsService.updateAffiliateBankPayment(donation)
+      totalDonated += donation.amount
+      updatedDonations++
+    }
+    if (trx.amount - totalDonated > 0 || updatedDonations < affiliate.donations.length) {
+      trx.bankDonationStatus = BankDonationStatus.incomplete
+      return ImportStatus.INCOMPLETE
+    }
```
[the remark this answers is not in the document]
sashko9807: All `incomplete` transactions, are now considered imported, if no guaranteed donations are found for the affiliate, on the next IRIS call. 

Is it ok that way??

apps/api/src/affiliate/affiliate.controller.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  @Post(':affiliateCode/donation')
+  @Public()
+  async createAffiliateDonation(
+    @Param('affiliateCode') affiliateCode: string,
+    @Body() donation: CreateAffiliateDonation,
+  ) {
+    const affiliate = await this.affiliateService.findOneByCode(affiliateCode)
+    if (!affiliate || !affiliate.company || !affiliate.company.person)
```
BogoCvetkov, line 86: Completely sane and valid condition. Could be reduced ro `if (!affiliate?.company?.person)`  -that's just informational

apps/api/src/affiliate/affiliate.controller.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  @Patch(':affiliateCode/donations/:donationId/cancel')
+  @Public()
+  async cancelAffiliateDonation(
+    @Param('affiliateCode') affiliateCode: string,
+    @Param('donationId') donationId: string,
+  ) {
+    const donation = await this.donationService.getAffiliateDonationById(donationId, affiliateCode)
+    if (!donation) throw new NotFoundException('Donation with this id is not found')
+
```
BogoCvetkov, line 123: After the donation has been confirmed by the iris cron - meaning money has been paid. Should we add a check to  prevent canceling in case the donation transitioned from guaranteed -> hower the final paid status is called

apps/api/src/auth/auth.service.ts
```diff
@@ -156,12 +156,15 @@ export class AuthService {
         await this.authenticateAdmin()
         const userData = await this.admin.users.findOne({ id: user.sub })
         const registerDto: RegisterDto = {
+          type: ProfileType.INDIVIDUAL,
           email: userData?.email ?? '',
           password: '',
           firstName: userData?.firstName ?? '',
           lastName: userData?.lastName ?? '',
+          companyName: 'typechecking-hack',
```
BogoCvetkov, line 164: Can we make the field optional, so that we won't have to add artificial data for non-company profiles

apps/api/src/tasks/bank-import/import-transactions.task.ts
```diff
@@ -401,7 +509,19 @@ export class IrisTasks {
...
     // Insert new transactions
-    const inserted = await this.prisma.bankTransaction.createMany({ data, skipDuplicates: true })
+    const inserted = await this.prisma.$transaction(
+      data.map((trx) =>
+        this.prisma.bankTransaction.upsert({
+          where: { id: trx.id },
+          update: {
+            bankDonationStatus: trx.description.startsWith(this.AFFILIATE_CODE_PREFIX)
+              ? trx.bankDonationStatus
+              : undefined,
```
BogoCvetkov, line 519: wouldn't this set the status of existing non-affiliate donation to undefined, which would be a problem if it was marked imported already

sashko9807: I will test it out just to be certain, but as per prisma's documentation `undefined` means **do nothing/ don't update** 
[Reference](https://www.prisma.io/docs/concepts/components/prisma-client/null-and-undefined)

apps/api/src/config/validation.config.ts
```diff
@@ -9,6 +9,7 @@ const globalValidationPipe = new ValidationPipe({
   transformOptions: {
     strategy: 'exposeAll',
     excludeExtraneousValues: true,
+    exposeDefaultValues: true,
```
BogoCvetkov, line 12: what is the purpuse of this

apps/api/src/tasks/bank-import/import-transactions.task.ts
BogoCvetkov: You are right. Than this is a completely fine

apps/api/src/config/validation.config.ts
sashko9807: If a field inside the dto is initialized , the field with the initialized value will be "attached" to the request body, if that field was not provided in the request.

Without setting this property to true, the initialized field from [CreateAffiliateDonationDto](https://github.com/podkrepi-bg/api/pull/564/files#diff-f6b5c69a98f12a5d976a182325ca630f28c638f37d797bb95194f8b19a9397f4) will be undefined, if not provided.

Though looking at it more clearly now, I should remove the validator decorators from `affiliateId` and leave it just as a type, as it something we should receive from the affiliateCode, rather than it being part of the request.

apps/api/src/affiliate/affiliate.controller.ts
sashko9807: The check is done through calling `shouldAllowStatusChange(donation,status, 'cancelled').

apps/api/src/auth/auth.service.ts
sashko9807: Technically, those fields are not added anywhere, as calling `createPerson`, doesn't address companyName, and companyNumber as part of the `person` update. Though changing those fields in `RegisterDto` to be `string | undefined`,  seems more appropriate, so will do.

apps/api/src/affiliate/affiliate.controller.ts
BogoCvetkov: Nice

apps/api/src/auth/auth.service.ts
BogoCvetkov: Perfect

apps/api/src/config/validation.config.ts
BogoCvetkov: Yes, you are right. BEcause you get the affiliate document from the code. When you say initialized do you mean the fields where you use this pattern `string | null = null`. My only concern here is since the option is a global and affects all dto's, that  it won't create unexpected results in other routes. But I suppose this is the first time this initialization pattern is being used, so should not be a problem

apps/api/src/donations/donations.controller.ts
```diff
@@ -175,7 +175,14 @@ export class DonationsController {
 
   @Get('user/:id')
   async userDonationById(@Param('id') id: string, @AuthenticatedUser() user: KeycloakTokenParsed) {
-    return await this.donationsService.getUserDonationById(id, user.sub, user.email)
+    const donation = await this.donationsService.getUserDonationById(id, user.sub, user.email)
+    return {
+      ...donation,
+      person: {
+        firstName: user.given_name,
+        lastName: user.family_name,
```
igoychev, line 183: not a big problem, but still the user names should be already available in the session on the frontend

apps/api/src/auth/auth.service.ts
```diff
@@ -352,6 +354,12 @@ export class AuthService {
 
   async changeEnabledStatus(keycloakId: string, enabled: boolean) {
     await this.authenticateAdmin()
+    // check if user is admin before attempting to activate/deactivate
+    const userGroups = await this.admin.users.listRoleMappings({ id: keycloakId })
+    const isAdmin = userGroups.realmMappings?.some((obj) => obj.name === 'podkrepi-admin')
```
igoychev, line 359: who is setting this role? so far we were using the ViewSupporters role for that

apps/api/src/donations/dto/donation-metadata.dto.ts
```diff
@@ -12,7 +16,7 @@ export class DonationMetadataDto {
   @ApiProperty()
   @Expose()
   @IsObject()
+  @Type(() => TExtraData)
   @IsOptional()
-  //eslint-disable-next-line
-  extraData: any
+  extraData: TExtraData | undefined
```
igoychev, line 21: might be good to add an example here of the expected structure, because my old mind fails to compile it ;)

igoychev (approved):

PR #565, opened 2023-10-19
Added campaign statistics endpoints

tongo-angelov:
Created a few endpoints needed for displaying charts in campaign's page

Since prisma does not have implementation for DATE_TRUNC, had to use queryRaw

### New endpoints - /statistics

- **/donations** - list campaign's donations grouped by either day, week or month
- **/unique-donations** - group donations with the same amount
- **/hourly-donations** - group donations by the time of day


apps/api/src/statistics/statistics.service.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  async listHourlyDonations(campaignId: string): Promise<HourlyDonationsDto[]> {
+    return this.prisma.$queryRaw`
+    SELECT EXTRACT(HOUR from created_at)::INTEGER AS hour, COUNT(id)::INTEGER AS count
+    FROM api.donations where status = 'succeeded'
+    ${Prisma.sql`AND target_vault_id IN ( SELECT id from api.vaults WHERE campaign_id = ${campaignId}::uuid)`}
+    GROUP BY hour
+    ORDER BY hour ASC`
+  }
```
igoychev, line 58: just wondering if GROUP BY DATE_TRUNC('HOUR') could work for the hourly donations so that it could be implemented inside listGroupedDonations too

tongo-angelov: DATE_TRUNC returns the nearest hour, so donations in the same time of day but different day will show different datetime (e.g:  2023-11-07T10:00:00.000Z , 2023-11-08T10:00:00.000Z)
Where by extracting the hour from the datetime we can group those given at the same time of day

igoychev: I see, and yes it is better in the way you did it!

igoychev (approved):
All good, well done!

Issue #515, opened 2023-07-12
Validate file types on upload


yg1y:
### Describe the feature you'd like to request

There are several screens through which files can be uploaded. There is no limit to the types of files that can be uploaded, and potentially dangerous files can be uploaded. It would be nice to have a check on these types of files that are allowed to be uploaded.

### Describe the solution you'd like

Allow only certain file types to be uploaded like .pdf, docx, .png, .jpeg, .xlsx ... and others

PR #573, opened 2023-11-03
Validation of file types

Nnachevvv:
Closes https://github.com/podkrepi-bg/api/issues/515

This PR provides mechanism to validate the file type uploaded to the backend. I added two types of validation , since using only the extension is not secure enough. Now we will support the following extension and mime types:

```typescript
  const allowedExtensions = /txt|json|jpeg|jpg|png|xml|xlsx|xls|docx/
  const mimeAllowlist = [
    'text/plain',
    'application/json',
    'application/pdf',
    'image/png',
    'image/jpeg',
    'application/xml',
    'text/xml',
    'application/msword',
    'application/vnd.ms-excel',
    'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
  ]
```

## Testing
Tried to upload one valid and invalid file:
Screenshot 2023-11-03 at 21 40 24
Failed with:
Screenshot 2023-11-03 at 21 40 45

Allowed mime type and not allowed extension:
Screenshot 2023-11-03 at 21 49 54

Screenshot 2023-11-03 at 21 50 04

### DIFF
--- a/apps/api/src/common/files.ts
+++ b/apps/api/src/common/files.ts
@@ -24,6 +24,7 @@ export function validateFileType(
   const mimeAllowlist = [
     'text/plain',
     'application/json',
+    'application/pdf',
     'image/png',
     'image/jpeg',
     'application/xml',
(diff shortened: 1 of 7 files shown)


apps/api/src/common/files.ts
```diff
@@ -0,0 +1,47 @@
...
+    'text/plain',
+    'application/json',
+    'image/png',
+    'image/jpeg',
+    'application/xml',
+    'text/xml',
+    'application/msword',
+    'application/vnd.ms-excel',
+    'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
+  ]
```
igoychev, line 34: let's add a pdf as allowed too

Nnachevvv: Added

igoychev (approved):
Looks great!

PR #574, opened 2023-11-05
Made currency required in the API, cleaned up the create or connect

yyosifov:
* Set the `currency` as required when creating a new Campaign via the API
* Cleaned up the vault resolution logic (create or connect) - there is always a default vault created when a Campaign is created

Closes 420


igoychev (approved):
Looks great! 
Well done with ensuring there will always be a default vault when creating the campaign! findFirstOrThrow!

PR #579, opened 2023-11-21
Feature/permanent delete person

PetarDimitrov91:
Added functionality that is responsible for deleting a user account. There is a safeguard in place that prevents the deletion of corporate, beneficiary, or organizer profiles. Additionally, several tests were added to ensure the proper functioning of this feature.


apps/api/src/auth/auth.service.ts
```diff
@@ -426,4 +428,25 @@ export class AuthService {
           )
     }
   }
+
+  async deleteUser(keycloakId: string) {
+    const user = await this.personService.findOneByKeycloakId(keycloakId)
+
+    //Check and throw if user is a beneficiary, organizer or corporate profile
+    if (user?.beneficiaries?.length || user?.organizer || user?.companyId) {
```
slavcho, line 436: I would change this to:

   `if (user?.beneficiaries?.length > 0`

Seems more natural to read.

apps/api/src/auth/auth.service.ts
```diff
@@ -426,4 +428,25 @@ export class AuthService {
...
+      throw new InternalServerErrorException(
+        'Cannot delete a beneficiary, organizer or corporate profile',
+      )
+    }
+
+    return this.authenticateAdmin()
```
slavcho, line 442: Personally, I am fine with this .then format.
But I think so far the common pattern is to use await.

slavcho (approved):
Just minor suggestions.

Looks good to me.
Love the unit tests.

apps/api/src/auth/auth.service.ts
PetarDimitrov91: I will adjust this later today.

apps/api/src/auth/auth.service.ts
PetarDimitrov91: I made this decision because if, for some reason, the promise from Keycloak does not resolve successfully, I want to interrupt the chain and refrain from removing the user in the API database. I am unsure if this is possible when I use await for every promise.

PR #581, opened 2023-12-07
src/donations: Fix wrong unique donors count

sashko9807:
Currently when we count unique donors we group them by billing_name, which gives us the total number of unique donors. The number of total donors is then added, to the total amount of donations made by the first user in the returned query. This results in overestimating the actual count of unique donors.

Changed the query to plain SQL, as prisma doesn't yet support counting by distinct properties, and counting is relatively faster to process, than grouping or searching 

Changed the query to count by distinct billing_email property, rather than billing_name, as email is the "more unique" property of the two.

### DIFF
--- a/apps/api/src/donations/donations.service.ts
+++ b/apps/api/src/donations/donations.service.ts
@@ -770,22 +770,14 @@ export class DonationsService {
   }
 
   async getDonorsCount() {
-    const donorsCount = await this.prisma.donation.groupBy({
-      by: ['billingName'],
-      where: { status: DonationStatus.succeeded },
-      _count: {
-        _all: true,
-      },
-      orderBy: { billingName: { sort: 'asc', nulls: 'first' } },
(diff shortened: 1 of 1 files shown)


apps/api/src/donations/donations.service.ts
```diff
@@ -770,22 +770,14 @@ export class DonationsService {
...
-      where: { status: DonationStatus.succeeded },
-      _count: {
-        _all: true,
-      },
-      orderBy: { billingName: { sort: 'asc', nulls: 'first' } },
-    })
-
-    // get count of the donations with billingName == null
-    const anonymousDonations = donorsCount[0]._count._all
+    const donorsCount = await this.prisma.$queryRaw<{
```
slavcho, line 773: queryRaw - this is usefull to learn :)

slavcho (approved):

PR #582, opened 2023-12-10
src/withdrawal: Allow deletion of withdrawal records

sashko9807:
When withdrawal record is deleted decrement blockedAmount of the vault, with the amount of the withdrawal

Closes: Reported in discord

### DIFF
--- a/apps/api/src/withdrawal/withdrawal.controller.spec.ts
+++ b/apps/api/src/withdrawal/withdrawal.controller.spec.ts
@@ -1,4 +1,4 @@
-import { NotFoundException, ForbiddenException, BadRequestException } from '@nestjs/common'
+import { NotFoundException, BadRequestException } from '@nestjs/common'
 import { Test, TestingModule } from '@nestjs/testing'
 import { WithdrawStatus, Currency } from '@prisma/client'
 import { mockReset } from 'jest-mock-extended'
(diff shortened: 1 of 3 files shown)


apps/api/src/withdrawal/withdrawal.controller.spec.ts
```diff
@@ -362,9 +362,37 @@ describe('WithdrawalController', () => {
...
+      prismaMock.withdrawal.delete.mockResolvedValue(withdrawal)
+      await expect(controller.remove(withdrawal.id)).toResolve()
+      expect(prismaMock.withdrawal.delete).toHaveBeenCalledWith({
+        where: { id: withdrawal.id },
+      })
+      expect(prismaMock.vault.update).toHaveBeenCalledWith({
+        where: { id: withdrawal.sourceVaultId },
+        data: {
+          amount: {
+            increment: withdrawal.status === WithdrawStatus.succeeded ? withdrawal.amount : 0,
```
slavcho, line 376: why do we have this status condition in the unit test?
Does this fluctuate?

apps/api/src/withdrawal/withdrawal.service.ts
```diff
@@ -124,11 +118,25 @@ export class WithdrawalService {
...
-    const result = await this.prisma.withdrawal.delete({ where: { id: id } })
-    if (!result) throw new NotFoundException('Not found')
+    const result = await this.prisma.withdrawal
+      .delete({
+        where: { id: id },
+      })
+      .catch(() => {
+        throw new BadRequestException("Withdrawal record couldn't be deleted")
+      })
+
```
slavcho, line 129: I think it is better to check the status before updating, instead of incrementing with 0...

slavcho (changes requested):

slavcho (approved):
I think I understand now...

PR #583, opened 2023-12-11
Update Notification templates

BogoCvetkov:
Update the fields for the Notification templates

### DIFF
--- a/apps/api/src/campaign-news/campaign-news.service.ts
+++ b/apps/api/src/campaign-news/campaign-news.service.ts
@@ -101,7 +101,7 @@ export class CampaignNewsService {
         campaignid: campaign.id,
         template_data: {
           'campaign.name': campaign?.title,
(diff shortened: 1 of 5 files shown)


apps/api/src/campaign-news/campaign-news.service.ts
```diff
@@ -101,7 +101,7 @@ export class CampaignNewsService {
         campaignid: campaign.id,
         template_data: {
           'campaign.name': campaign?.title,
-          'campaign.target-amount': campaign?.targetAmount || 0,
+          'campaign.target-amount': (campaign?.targetAmount && campaign?.targetAmount / 100) || 0,
```
slavcho, line 104: This is somewhat difficult to read. 
I wonder if there is a better way to get the campaign target.

slavcho (approved):
Looks good to me.

slavcho:
There are a couple of unit tests failing. 
I think we should fix those before merging...

PR #584, opened 2023-12-12
src/vault: Include campaign data and sum of succeeded withdraws to vault's response

sashko9807:
Needed for better front-end visualization


apps/api/src/vault/vault.service.ts
```diff
@@ -25,8 +31,33 @@ export class VaultService {
     return await this.prisma.vault.create({ data: createVaultDto.toEntity() })
   }
 
-  async findAll(): Promise<Vault[]> {
-    return await this.prisma.vault.findMany()
+  async findAll(): Promise<VaultWithWithdrawalSum[]> {
```
slavcho, line 34: Wow! That is some SQL magic!

slavcho (approved):

PR #589, opened 2023-12-21
src/donations: Allow status update from status declined

sashko9807:
For some reason some donations from stripe are first with status declined, which later is updated to succeeded, but which we can't currently update.


slavcho (approved):
Looks good to me.
Seems like stripe made a change in the protocol last month.

PR #597, opened 2024-01-16
Invalidate a successfull donation

slavcho:
This invalidates a successful donation and removes the money from the vault.

## Motivation and context
In some rare cases we have a duplicated or an invalid donation from stripe.
We want to be able to manual adjust this from the admin panel.

We would need to add a new role for that in keycloak: account-edit-financials-requests

## Testing
Provided a unit test and tested manually.

### DIFF
--- a/apps/api/src/donations/donations.controller.ts
+++ b/apps/api/src/donations/donations.controller.ts
@@ -16,7 +16,7 @@ import {
 import { ApiQuery, ApiTags } from '@nestjs/swagger'
 import { DonationStatus } from '@prisma/client'
 import { AuthenticatedUser, Public, RoleMatchingMode, Roles } from 'nest-keycloak-connect'
-import { RealmViewSupporters, ViewSupporters } from '@podkrepi-bg/podkrepi-types'
+import { RealmViewSupporters, ViewSupporters, EditFinancialsRequests } from '@podkrepi-bg/podkrepi-types'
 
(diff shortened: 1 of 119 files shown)


apps/api/src/donations/donations.controller.ts
```diff
@@ -254,9 +254,19 @@ export class DonationsController {
     return this.donationsService.update(id, updatePaymentDto)
   }
 
+  @Post('/invalidate-stripe-payment/:id')
```
sashko9807, line 257: HTTP `Patch` seems  the more appropriate form of request for this use case.
The endpoint could also be changed to `/:id/invalidate`, as what is being invalidated is existing donation record in our DB, rather than stripe payment. Depending on the implementation, this could also be used to invalidate bank donations.

sashko9807 (changes requested):

apps/api/src/donations/donations.controller.ts
slavcho: I've tried that but it doesn't work.
There is another Patch( /: id) handler and any request to /:id/invalidate is captured by the first one.

sashko9807: Try putting the definition of @Patch (:id/invalidate ) handler above @Patch(:id) one. Should work that way.

slavcho: I tried that as well :)

sashko9807 (approved):

PR #599, opened 2024-01-22
src/donations: Don't update updatedAt field on iris re-import

sashko9807:
Since campaign donations are ordered by updatedAt field on the front-end, this causes to bank donations to reappear at top every hour. 
This is temporary solution until the reason for updating existing donation record is figured out.

### DIFF
--- a/apps/api/src/donations/donations.service.ts
+++ b/apps/api/src/donations/donations.service.ts
@@ -583,7 +583,7 @@ export class DonationsService {
       //Donation exists, so updating with incoming donation without increasing vault amounts
       await this.prisma.donation.update({
         where: { extPaymentIntentId: donationDto.extPaymentIntentId },
-        data: donationDto,
+        data: { ...donationDto, updatedAt: existingDonation.updatedAt },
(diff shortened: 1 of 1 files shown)


apps/api/src/donations/donations.service.ts
```diff
@@ -583,7 +583,7 @@ export class DonationsService {
       //Donation exists, so updating with incoming donation without increasing vault amounts
       await this.prisma.donation.update({
         where: { extPaymentIntentId: donationDto.extPaymentIntentId },
-        data: donationDto,
+        data: { ...donationDto, updatedAt: existingDonation.updatedAt },
```
slavcho, line 586: Just for my curiosity... 
if the donation already exists - why do we even update it?

sashko9807: I have no idea either thus decided to keep it for now, just in case of specific edge-case I can't think of. 

@BogoCvetkov is it safe to remove this update query?

slavcho (approved):

PR #602, opened 2024-02-12
more campaign info added in campaign-news service

RalitsaIlieva:
The change was made because I need more info about the campaign in the news section where a new frontend is prepared

---|---|---|---

### DIFF
--- a/apps/api/src/campaign/campaign.controller.ts
+++ b/apps/api/src/campaign/campaign.controller.ts
@@ -128,12 +128,9 @@ export class CampaignController {
   }
 
   @Get('byId/:id')
-  @Roles({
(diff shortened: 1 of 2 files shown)


apps/api/src/campaign/campaign.controller.ts
```diff
@@ -128,12 +128,9 @@ export class CampaignController {
   }
 
   @Get('byId/:id')
```
sashko9807, line 130: This endpoint is for administrators only, and the returned data by it is not meant to be public. 
If the goal is to get the donation data of campaign, please use one of the available Public() endpoints

sashko9807 (changes requested):

PR #608, opened 2024-03-13
Add a script that fixes duplicated donations on the database.

slavcho:
Those were caused by a bug in the bank imports and two donations have been multiplied several times.


sashko9807:
Closing this, as it is fixed.

PR #611, opened 2024-03-14
donations: Add endpoint which would allow to sync amount of donation and payment

sashko9807:
There are some donations with amount of 0. Allow to manually update them with payment's amount


slavcho (approved):
This should be used smartly. As it works only in case of one payment - one donation.

PR #612, opened 2024-03-15
src/donations: Fix donation filter query

sashko9807:
Fixes query filters not working for donation section on admin panel


apps/api/src/donations/donations.service.ts
```diff
@@ -385,12 +385,12 @@ export class DonationsService {
         lte: to,
       },
       paymentId: paymentId,
-      OR: [
-        { payment: { status: paymentStatus } },
-        { payment: { provider: paymentProvider } },
-        { payment: { billingEmail: { contains: search } } },
-        { payment: { billingName: { contains: search } } },
-      ],
+      payment: {
```
slavcho, line 388: doesn't this search with an AND now?
It used  to be OR...

slavcho (approved):

PR #613, opened 2024-03-16
src/donations: Set donation record date to payment's

sashko9807:
When syncing donation with payment, use payment's createdAt , updatedAt params. 
This is due to us currently using updatedAt field, to list donations to campaign.


sashko9807:
Closing this for now, in favour of better approach

PR #634, opened 2024-05-15
fix: Check if sendgrid list exists before attempting to create new one

sashko9807:
For some reason,  campaigns' email list exists on sendgrid, but the data is not stored in the internal notification_list table. This behavior results in Bad Request whenever user attempts to subscribe for email notifications for a particular campaign , as we try to create two lists with the same name.

Closes [frontend#1170](https://github.com/podkrepi-bg/frontend/issues/1770)

### DIFF
--- a/apps/api/src/campaign/campaign.service.ts
+++ b/apps/api/src/campaign/campaign.service.ts
@@ -1040,7 +1040,7 @@ export class CampaignService {
     let listId: string
     const lists = await this.marketingNotificationsService.provider.getContactLists()
     const campaginEmailLists = lists.body.result
-    const exists = campaginEmailLists.find((campaign) => campaign.name === updated.title)
+    const exists = campaignEmailLists.find((campaign) => campaign.name === updated.title)
     if (exists) {
       listId = exists.id
     } else {
--- a/apps/api/src/campaign/campaign.service.ts
+++ b/apps/api/src/campaign/campaign.service.ts
@@ -1037,7 +1037,7 @@ export class CampaignService {
 
   async createCampaignNotificationList(updated: { title: string; id: string }) {
     // Generate list in the marketing platform
-    let listId: string = ''
+    let listId: string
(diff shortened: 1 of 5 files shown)


apps/api/src/campaign/campaign.service.ts
```diff
@@ -1037,9 +1037,17 @@ export class CampaignService {
 
   async createCampaignNotificationList(updated: { title: string; id: string }) {
     // Generate list in the marketing platform
-    const listId = await this.marketingNotificationsService.provider.createNewContactList({
-      name: updated.title || updated.id,
-    })
+    let listId: string
+    const lists = await this.marketingNotificationsService.provider.getContactLists()
+    const campaginEmailLists = lists.body.result
+    const exists = campaginEmailLists.find((campaign) => campaign.name === updated.title)
```
imilchev, line 1043: nit: typo
```suggestion
    const exists = campaignEmailLists.find((campaign) => campaign.name === updated.title)
```

apps/api/src/campaign/campaign.service.ts
```diff
@@ -1037,9 +1037,17 @@ export class CampaignService {
...
+    if (exists) {
+      listId = exists.id
+    } else {
+      listId = await this.marketingNotificationsService.provider.createNewContactList({
```
imilchev, line 1047: do we want to silently create a list here? I would assume we should be only doing anything if a list already exists?

sashko9807: The current issue is that for some reason, email lists for some campaigns exist in Sendgrid database, but don't exist in ours NotificationList. This causes an attempt to create two email lists with the same name, which results in Bad Request whenever somebody attempts to subscribe for email notifications to particular campaign.

What this peace of code does is to check, whether there is already a Sendgrid list with the same name, before attempting to create a new one.

slavcho (approved):
Looks OK to me.

PR #639, opened 2024-06-06
fix: Donation wish not being created after successfull donation

sashko9807:
An issue was detected where, donation wishes aren't imported on charge.succeeded event.

### DIFF
--- a/apps/api/src/campaign/campaign.service.ts
+++ b/apps/api/src/campaign/campaign.service.ts
@@ -603,29 +603,28 @@ export class CampaignService {
     return await this.prisma.$transaction(async (tx) => {
       let donationId
       // Find donation by extPaymentIntentId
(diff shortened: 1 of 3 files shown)


apps/api/src/campaign/campaign.service.ts
```diff
@@ -603,29 +603,28 @@ export class CampaignService {
...
-      if (!existingDonation) {
+      if (!existingPayment) {
         const newDonation = await this.createIncomingDonation(
           tx,
           paymentData,
           newDonationStatus,
           campaign,
         )
-        donationId = newDonation.id
+        donationId = newDonation.donations[0].id
```
imilchev, line 616: is it always just 1 donation? can this be 0 or more than 1?

sashko9807: At this point it will always be 1 donation.

slavcho (approved):

PR #645, opened 2024-06-18
feat: Add endpoint to send newsletter consent email

sashko9807:
Adds an endpoint to send newsletter consent mails.

- Added a function to export contacts from Sendgrid list
- Added a function to chunk sendgrid contacts list
- Added a function to send mass emails which are scheduled in a minute difference.


ani-kalpachka (approved):

apps/api/src/notifications/notifications.service.ts
```diff
@@ -492,4 +509,88 @@ export class MarketingNotificationsService {
...
+      contacts.set(email, {
+        id: id,
+        hash: this.generateHash(id),
+        registered: false,
+      })
+    }
+  }
+
+  private updateMailListMap(
+    regUser: Person[],
```
slavcho, line 526: regUser -> regUsers

apps/api/src/notifications/notifications.service.ts
```diff
@@ -492,4 +509,88 @@ export class MarketingNotificationsService {
...
+    contacts: ContactsMap,
+    skipAfterDate: Date,
+    unregisteredConsent: UnregisteredNotificationConsent[],
```
slavcho, line 529: unregisteredConsent -> unregisteredConsents

apps/api/src/notifications/notifications.service.ts
```diff
@@ -492,4 +509,88 @@ export class MarketingNotificationsService {
...
+  ) {
+    for (const registeredUser of regUser) {
+      const createdAt = new Date(registeredUser.createdAt)
+
+      // Remove email if it belongs to user created after the change has been deployed, as they had already decided
+      // whether to give consent or not.
+      if (contacts.get(registeredUser.email as string) && createdAt > skipAfterDate) {
+        Logger.debug(`Removing email ${registeredUser.email} from list`)
```
slavcho, line 537: Can we add a bit more in the log:

removing email XXX from the list because of date

and same later below:

removing email XXX because the user don't want them.

apps/api/src/notifications/providers/notifications.sendgrid.provider.ts
```diff
@@ -39,6 +56,78 @@ export class SendGridNotificationsProvider
...
+    let numOfRetries = 0
+    Logger.debug('Creating contacts exports')
+    const createContactExport = await this.createContactExport(listId)
+    const jobId = createContactExport.id
+    Logger.debug(`Created export with id ${jobId}`)
+    let exportStatusResponse = await this.getContactExportStatus(jobId)
+
+    do {
+      Logger.debug('Waiting export to be finished')
+      await new Promise((r) => setTimeout(r, SENDGRID_EXPORT_TIMEOUT))
```
slavcho, line 97: I don't understand how this works... so we wait 10 seconds every time?
Seems very strange

apps/api/src/notifications/providers/notifications.sendgrid.provider.ts
```diff
@@ -39,6 +56,78 @@ export class SendGridNotificationsProvider
...
+        default:
+      }
+      numOfRetries++
+    } while (exportStatusResponse.status === 'pending' && numOfRetries < RETRY_LIMIT)
+    if (numOfRetries >= RETRY_LIMIT) {
+      throw new InternalServerErrorException(
+        `Couldn't export contacts within the limit. Try again later.`,
+      )
+    }
+    const exportUrl = exportStatusResponse.urls[0]
```
slavcho, line 114: can we check if the urls list is non empty?

apps/api/src/notifications/providers/notifications.sendgrid.provider.ts
```diff
@@ -39,6 +56,78 @@ export class SendGridNotificationsProvider
...
+    const response = await fetch(exportUrl, {
+      headers: { 'Content-Type': 'application/json', 'Accept-Encoding': 'zlib' },
+    })
+
+    const gzipped = await response.arrayBuffer()
+    const buffer = Buffer.from(gzipped)
```
slavcho, line 120: So we ask (and expect) for gzipped response. 
Does "Buffer" unzipped it somehow?

apps/api/src/notifications/providers/notifications.sendgrid.provider.ts
```diff
@@ -180,4 +269,53 @@ export class SendGridNotificationsProvider
 
     return response
   }
+
+  async sendBulkEmail(data: MassMailDto, contactsMap: ContactsMap[], value: string): Promise<void> {
+    const currentDate = new Date()
+    contactsMap.forEach((contacts, index) => {
+      //Schedule  batches in a minute difference
+      currentDate.setMinutes(currentDate.getMinutes() + index)
```
slavcho, line 277: This will result in current date with the following values (say it is now 11:00):

11:00
11:01 
11:03
11:06
11:10
11:15
11:21
...

If I am reading this correctly.
It will probably work, I am just not sure this was the intent, especially if we expect this list to be big...

apps/api/src/notifications/providers/notifications.sendgrid.provider.ts
sashko9807: Sendgrid doesn't have API to extract emails from list as of now, thus we need first to create export file, which takes approx. ~10-15 seconds.

apps/api/src/notifications/providers/notifications.sendgrid.provider.ts
sashko9807: As far as i recall the response is not compressed. The naming was from an older implementation. Will rename it to prevent confusion

apps/api/src/notifications/providers/notifications.sendgrid.provider.ts
sashko9807: Will do, though if export status is succeeded it always returns an url.

apps/api/src/notifications/providers/notifications.sendgrid.provider.ts
sashko9807: Will have to revisit whether the scheduling part works as intended , but the idea is to schedule each call in a minute difference, due to Sendgrid Rate limiting. eg:
First batch of emails: 10:00
Second batch of emails: 10:01
Third batch of emails: 10:02

Big lists are not a problem since those emails are sent in bulk. What the implementation does is the following:
Get contacts data from list -> Split contact's data into chunks, with each chunk having up to 1000 emails -> Schedule mail send for the emails within each chunk.

Though thinking about it we should probably have a way to cancel the mail sent at some point.

apps/api/src/notifications/providers/notifications.sendgrid.provider.ts
slavcho: I see

apps/api/src/notifications/providers/notifications.sendgrid.provider.ts
sashko9807: Done.

apps/api/src/notifications/providers/notifications.sendgrid.provider.ts
sashko9807: Lets test this in staging/prod environment? We already have a private contact list, to test this internally first?

slavcho (approved):

PR #648, opened 2024-06-23
chore: campaign application create contract

gparlakov:
- expose the create-application endpoints to the swagger docs
- flesh out the create campaign data transfer object as per https://github.com/users/gparlakov/projects/1/views/1?pane=issue&itemId=68403341


sashko9807 (approved):
Looks good.
Just for reference `@AuthenticatedUser` decorator, is only needed if we want to extract data from the AccessToken(e.g. keycloakId of the user making the request) which will be used for something. Otherwise it is not necessary.

PR #649, opened 2024-06-25
add create campaign agreements  validation

Martbul:
if any of the 3 agreements for creating campaign are not checked -> throw a HttpException with status 405

add validation if the 3 agreements are checked(true)
add test
update create-campaign-application.dto(acceptTermsAndConditions, transparencyTermsAccepted,personalInformationProcessingAccepted are boolean)

---|---|---|---

### DIFF
--- a/apps/api/src/campaign-application/campaign-application.service.ts
+++ b/apps/api/src/campaign-application/campaign-application.service.ts
@@ -1,4 +1,4 @@
-import { HttpException, HttpStatus, Injectable } from '@nestjs/common'
+import { BadRequestException, HttpException, HttpStatus, Injectable } from '@nestjs/common'
 import { CreateCampaignApplicationDto } from './dto/create-campaign-application.dto'
 import { UpdateCampaignApplicationDto } from './dto/update-campaign-application.dto'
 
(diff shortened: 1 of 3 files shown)


apps/api/src/campaign-application/campaign-application.service.ts
```diff
@@ -9,6 +9,13 @@ export class CampaignApplicationService {
   }
 
   create(createCampaignApplicationDto: CreateCampaignApplicationDto) {
+    if (
+      !createCampaignApplicationDto.acceptTermsAndConditions ||
+      !createCampaignApplicationDto.transparencyTermsAccepted ||
+      !createCampaignApplicationDto.personalInformationProcessingAccepted
+    ) {
+      throw new HttpException('All agreements must be checked', HttpStatus.METHOD_NOT_ALLOWED)
```
sashko9807, line 17: `Method not allowed (405)` is not the appropriate http response to respond with in such scenarios. 
We've established some practice to respond with  `Bad Request (400)` for invalid DTO content, but I suppose you could also respond with `Unprocessable Content (422)`. Either of them are semantically fine.

sashko9807 (changes requested):

sashko9807 (approved):

PR #652, opened 2024-06-27
add findAll(GET) method in the campaign-application.service(second try)

Martbul:
Closes #{issue number}

---|---|---|---


apps/api/src/campaign-application/campaign-application.service.spec.ts
```diff
@@ -2,14 +2,57 @@ import { Test, TestingModule } from '@nestjs/testing'
...
-      providers: [CampaignApplicationService],
-    }).compile()
+      providers: [
+        {
+          // Use the interface as token
+          provide: NotificationsProviderInterface,
+          // But actually provide the service that implements the interface
+          useClass: SendGridNotificationsProvider,
+        },
+        CampaignService,
```
gparlakov, line 36: Do we need all these providers? CampaignService? Vault Service.... and all the rest? 
It looks like `CampaignApplicationService` is the only requirement for this PR's scope

gparlakov: Keep in mind that all of these would need to be instantiated with all of their dependencies for each test!!! 

Instead use mocking (see other tests or look for autoSpy usage)

apps/api/src/campaign-application/campaign-application.service.ts
```diff
@@ -20,7 +22,7 @@ export class CampaignApplicationService {
   }
 
   findAll() {
-    return `This action returns all campaignApplication`
+    return this.prisma.campaignApplication.findMany()
```
gparlakov, line 25: The controller would need to be updated to check for authenticated user and allow getting only for admin/operator roles.

gparlakov (commented):
needs a bit more ❤️

apps/api/src/campaign-application/campaign-application.service.spec.ts
gparlakov, line 1: Please remove commented out code

gparlakov (commented):
Thanks for updating.

apps/api/src/campaign-application/campaign-application.service.spec.ts
Martbul: Sure. I have missed it.

Martbul:
Трябва ли да правя някакви промени?

PR #653, opened 2024-06-28
add postman collection

gparlakov:
## Motivation and context

- plus documentation and examples
- the goal is to have new contributors be able to quickly test the api

## Testing
open with postman compatible tool (postman app, newman cli,...)
send requests


sashko9807 (commented):
Great contribution.
Though I suspect this will be used way more in the local environment, than the staging one. What do you think of setting another collection for the local environment?

The data can be prefilled with by default:
url: `http://localhost:5010`
user:`admin@podkrepi.bg`
password:`$ecurePa33`

gparlakov:
@sashko9807 agreed and updated

sashko9807 (approved):

PR #654, opened 2024-07-01
fix(seed.ts): Fix seeding script failing due to unique constraint

sashko9807:
Quick solution to fix seeding script


ani-kalpachka (approved):

slavcho (approved):
So now we create a wish for one donation instead...
Can you add a TODO comment to add for many some day?

PR #656, opened 2024-07-02
finish create functionality and testing on it

Martbul:
update CampaignApplication model - organizerEmail now is not unique 
update create method testing(cotroller and service file)
add create method functionality - now created campaign-application data is persisted to db
update createCampaignApplicationDto - add @isNotEmpty decorator for properties that are required

Closes #{issue number}

---|---|---|---


apps/api/src/campaign-application/campaign-application.service.spec.ts
```diff
@@ -1,83 +1,267 @@
...
+    history: 'Test history',
+    amount: '1000',
+    description: 'Test description',
+    campaignGuarantee: 'Test guarantee',
+    otherFinanceSources: 'Test otherFinanceSources',
+    otherNotes: 'Test otherNotes',
+    category: CampaignTypeCategory.medical,
+  }
+
+  const mockCampaigns = [
```
sashko9807, line 54: Can you create `__mocks__` folder and have one campaignMocks file in it, and export mockCampaigns from it?

apps/api/src/campaign-application/campaign-application.controller.ts
```diff
@@ -1,25 +1,53 @@
...
 import { KeycloakTokenParsed, isAdmin } from '../auth/keycloak'
+import { PersonService } from '../person/person.service'
 
 @ApiTags('campaign-application')
 @Controller('campaign-application')
 export class CampaignApplicationController {
-  constructor(private readonly campaignApplicationService: CampaignApplicationService) {}
+  constructor(
+    private readonly campaignApplicationService: CampaignApplicationService,
+    @Inject(forwardRef(() => PersonService)) private readonly personService: PersonService,
```
sashko9807, line 28: `@Inject` won't be needed if we import the PersonModule, to the `CampaignApplicationModule`. Please refer to one of the other comments from more information.

sashko9807 (changes requested):

apps/api/src/campaign-application/campaign-application.service.ts
```diff
@@ -1,26 +1,75 @@
...
+
+      let organizer = await this.prisma.organizer.findUnique({
+        where: { personId: person.id },
+      })
+
+      if (!organizer) {
+        organizer = await this.organizerService.create({
+          personId: person.id,
+        })
+      }
```
gparlakov, line 40: Not sure if we should create a person here. I'd say - if the organizer is missing - no new campaignApplication.

apps/api/src/campaign-application/campaign-application.service.ts
```diff
@@ -1,26 +1,75 @@
...
+        organizerId: organizer.id,
+      }
+
+      const newCampaignApplication = await this.prisma.campaignApplication.create({
+        data: sanitizedData,
+      })
+
+      return newCampaignApplication
+    } catch (error) {
+      console.error('Error in create():', error)
```
gparlakov, line 66: It looks like `Logger` (imported from 'nestjs`) is the standard logging used in other services. Let's use that. 

@sashko9807 since I'm kinda new in nestjs and looking at for example CampaignService.createCampaign - there is no explicit logging if create fails. So the question is - do we need to explicitly log or the Logger will log [class][method] Error message, error stack automatically if a prisma error appears?

apps/api/src/campaign-application/campaign-application.service.ts
Martbul: Here I am not creating person. Initialy there are no organizers, so when a person creates a new campaignApplication he becomes a organizer => creating organizer with the person's id

Martbul:
I commited to the wring branch, that updated the PR with wrong changes. So I closed the PR and made a new one. Sorry, I am still learning.

PR #657, opened 2024-07-03
Persisting created new campaignApplication to db

Martbul:


apps/api/src/campaign-application/campaign-application.controller.spec.ts
```diff
@@ -26,63 +42,61 @@ describe('CampaignApplicationController', () => {
     expect(controller).toBeDefined()
   })
 
-  it('when create called it should delegate to the service create', () => {
```
gparlakov, line 29: This is deleted but the logic is still in use. Was this by design?

apps/api/src/campaign-application/campaign-application.controller.spec.ts
```diff
@@ -2,21 +2,37 @@ import { Test, TestingModule } from '@nestjs/testing'
 import { CampaignApplicationController } from './campaign-application.controller'
 import { CampaignApplicationService } from './campaign-application.service'
 import { SpyOf, autoSpy } from '@podkrepi-bg/testing'
```
gparlakov, line 4: Some unuses stuff that github(or an action?) points out - SpyOf and CamapignTypeCategroy

apps/api/src/campaign-application/campaign-application.controller.spec.ts
```diff
@@ -26,63 +42,61 @@ describe('CampaignApplicationController', () => {
...
   })
 
   it('when findOne called it should delegate to the service findOne', () => {
-    // arrange
-    // act
+    // Act
     controller.findOne('id')
 
-    // assert
+    // Assert
```
gparlakov, line 91: avoid such unnecessary changes - they make the code review that much harder

apps/api/src/campaign-application/campaign-application.service.spec.ts
```diff
@@ -20,131 +41,117 @@ describe('CampaignApplicationService', () => {
...
+
+      expect(result).toEqual(mockCreatedCampaignApplication)
+
+      expect(mockOrganizerService.create).toHaveBeenCalledWith({
+        personId: mockPerson.id,
+      })
+
+      expect(prismaMock.campaignApplication.create).toHaveBeenCalledWith({
+        data: {
+          campaignName: dto.campaignName.trim(),
```
gparlakov, line 130: This looks like the logic itself and while it's technically correct I belive we'll get more value having hardcoded value here i.e.  {

data :{
campaignName: 'my-campaign',
organizerName: 'test organizer',

...

apps/api/src/campaign-application/campaign-application.service.ts
```diff
@@ -1,24 +1,64 @@
...
+      })
+
+      if (!organizer) {
+        organizer = await this.organizerService.create({
+          personId: person.id,
+        })
+      }
+
+      const sanitizedData = {
+        campaignName: createCampaignApplicationDto.campaignName.trim(),
```
gparlakov, line 36: @sashko9807 is there a sanitizing logic somewhere on the stack (nest, middleware, prisma perhaps)  that prevents Cross Site Scripting or SQL injection (i.e. my name is <script>alert('pawned')</script> or the famous [jonny drop tables](https://stackoverflow.com/questions/332365/how-does-the-sql-injection-from-the-bobby-tables-xkcd-comic-work)

sashko9807: Prisma sanitizes data on insertion. Although I am not sure whether the start-end spaces are automatically trimmed.  Don't recall seeing this practice in any other part of the codebase

Will have to check for that.

apps/api/src/campaign-application/campaign-application.service.ts
```diff
@@ -1,24 +1,64 @@
...
+        where: { personId: person.id },
+      })
+
+      if (!organizer) {
+        organizer = await this.organizerService.create({
+          personId: person.id,
+        })
+      }
+
+      const sanitizedData = {
```
sashko9807, line 35: Prisma sanitizes the input by default. Seems like it also trims unnecessary spaces from user input, thus `sanitizedData` is not needed.

sashko9807 (changes requested):

sashko9807 (changes requested):

sashko9807 (approved):

PR #661, opened 2024-08-14
add findAll and findOne for admin only usage

Martbul:


apps/api/src/campaign-application/campaign-application.controller.ts
```diff
@@ -76,8 +77,36 @@ export class CampaignApplicationController {
...
+    if (!person) {
+      Logger.error('No person found in database')
+      throw new NotFoundException('No person found in database')
+    }
+    let isAdminFlag
+    if (!isAdmin(user)) {
+      isAdminFlag = false
+    } else {
+      isAdminFlag = true
+    }
```
sashko9807, line 91: No need for the isAdminFlag. isAdmin already returns a boolean.
You can just do `const isAdmin = isAdmin(user)` and pass isAdmin to the the findOne service.

apps/api/src/campaign-application/campaign-application.controller.ts
```diff
@@ -76,8 +77,36 @@ export class CampaignApplicationController {
...
+
+  @Delete('fileById/:id')
+  async deleteFile(@Param('id') id: string, @AuthenticatedUser() user: KeycloakTokenParsed) {
+    const person = await this.personService.findOneByKeycloakId(user.sub)
+    if (!person) {
+      Logger.error('No person found in database')
+      throw new NotFoundException('No person found in database')
+    }
+    let isAdminFlag
+    if (!isAdmin(user)) {
```
sashko9807, line 103: Same as above

sashko9807 (changes requested):

apps/api/src/campaign-application/campaign-application.controller.spec.ts
```diff
@@ -18,6 +18,7 @@ describe('CampaignApplicationController', () => {
   let controller: CampaignApplicationController
   let service: CampaignApplicationService
   let personService: PersonService
+  const { isAdmin } = require('../auth/keycloak')
```
sashko9807, line 21: Use the ES6 import  `import {isAdmin} from `../auth/keycloak`, over the require() statement

sashko9807 (changes requested):

apps/api/src/campaign-application/campaign-application.service.ts
```diff
@@ -86,12 +86,66 @@ export class CampaignApplicationService {
...
+    try {
+      const campaignApplications = await this.prisma.campaignApplication.findMany()
+      return campaignApplications
+    } catch (error) {
+      Logger.error('Error in findAll():', error)
+      throw error
+    }
+  }
+
+  async findOne(id: string, isAdminFlag: boolean, person: Person) {
```
sashko9807, line 99: A bit ugly but set the type of the person to be 
```js
person: Prisma.PersonGetPayload<{ include: { organizer: {select:{id:true}}}}>
```

instead of `Person` to prevent the linter complaint. Also make sure that organizer.id is optional(e.g. person.organizer?.id)

apps/api/src/campaign-application/campaign-application.service.ts
```diff
@@ -86,12 +86,66 @@ export class CampaignApplicationService {
...
+      return singleCampaignApplication
+    } catch (error) {
+      Logger.error('Error in findOne():', error)
+      throw error
+    }
   }
 
-  findOne(id: string) {
-    return `This action returns a #${id} campaignApplication`
+  async deleteFile(id: string, isAdminFlag: boolean, person: Person) {
```
sashko9807, line 119: same as above.

PR #669, opened 2024-09-07
fix: Sendgrid can't create list if name is too long

sashko9807:
Sendgrid's contact list name limit, is towards bytes, not a num of characters.

Closes [#1914](https://github.com/podkrepi-bg/frontend/issues/1914)


apps/api/src/notifications/helpers/truncateNameToBytes.ts
```diff
@@ -0,0 +1,21 @@
+export function truncateNameToBytes(name: string, maxBytes: number) {
+  const encoder = new TextEncoder()
```
slavcho, line 2: can we add some comments about what do those Encoders do?
I guess this does something like utf8 to bytes... and whole play is not to cut in the middle of a symbol.

slavcho (approved):

PR #678, opened 2024-11-24
fix: Connect non anonymous donation by personId

sashko9807:
It is not necessary for the billlingEmail to be the one of the registered user(e.g. company emails for billing)
Fixed website donation message data via websocket


gparlakov (approved):
Lgtm

PR #680, opened 2024-12-09
fix: Generate Stripe idempotency keys server-side

sashko9807:


apps/api/src/stripe/stripe.service.ts
```diff
@@ -220,19 +208,23 @@ export class StripeService {
...
     })
 
-    if (productLookup) return productLookup.data[0]
+    if (productLookup.data.length) return productLookup.data[0]
     return await this.stripeClient.products.create(
       {
         name: campaign.title,
         description: `Donate to ${campaign.title}`,
+        metadata: {
+          campaignId: campaign.id,
```
gparlakov, line 226: Do we need to do a migration for the existing products that were created without `campaign.id` ?

gparlakov (approved):

